PR #6, opened 2020-09-29
Setup CI for main branch

Everest2020:
In order to keep the project healthy, we would like to setup CI.
- Use dotnet CLI to build the solution.
- Use dotnet CLI to run the tests.


SukruthGrandhiNI (approved):

jryckman (commented):
Is there a way to add a stage to this pipeline that runs all the test in the test assemblies?

Everest2020:
> Is there a way to add a stage to this pipeline that runs all the test in the test assemblies?

I already added one, 

>   - name: Run Tests
        run: dotnet test NationalInstruments.Analyzers.sln

.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: windows-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+        name: Checkout code
```
jryckman: I don't fully understand yml, but it seems weird that the first usage here has `-uses`  first with the dash in front of it and name below, where everything else is the opposite. Is that intentional?

Everest2020: The order does not matter. According to the yaml syntax, the first section should begin with '-'

jryckman (approved):

PR #8, opened 2020-09-30
Added remaining rulesets, cleaned up stale configuration and analyzers and added documentation around banned methods

jryckman:
# Justification
This change does several things:
- Adds remaining rulesets that we want to be a part of our package
- Removes LiteralExemptions as it's a stale rule and needs to be removed from our analyzer
- Updates banned methods to include documentation on how to use it, but to also not include a preset one in our package. This is a better approach because banned methods should live in the consuming project's source control as opposed to our package, to mirror what we did with CodeAnalysisDictionary.xml
- Update NI.Analysis.targets to include NI.RoslynTransition.targets instead of having separate files, since it's not clear we're ever going to be able to remove it. I also added more TODO and documentation in the file to explain what is going on and what needs to change before we can package everything up.

# Implementation
I moved files around to be in slightly different locations (configuration directory), but that doesn't actually change anything until we package things up. For the particular path changes, I've updated spots in the NI.Analysis.targets file to point to the new, relative path.
There are some lines in the .targets file that were deleted and added, which I'll try to add comments around to identify the key changes vs. straight copy behavior.

# Testing
No testing this round since it's all configuration and no new analyzers. Relying on existing CI.


configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,107 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <!-- TODO: This location of 'content' needs to get updated once we build our package if we decide to put the files elsewhere, outside of a 'content' directory -->
```
jryckman, line 4: I added this TODO to note that we may need to update this once we create our new package. If it stays the same, we should just remove the TODO.

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,107 @@
...
+
+  <PropertyGroup>
+    <!-- Enable Roslyn analyzers if they haven't been explicitly disabled, we're using Roslyn for CA/SA, the code needs to be validated -->
+    <EnableRoslynAnalyzers>True</EnableRoslynAnalyzers>
+    <IncludeStyleAnalyzers>True</IncludeStyleAnalyzers>
+    <IncludeCodeAnalyzers>True</IncludeCodeAnalyzers>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <NI1006_BannedMethods>$(SolutionDir)\BannedMethods.xml</NI1006_BannedMethods>
```
jryckman, line 31: I removed Literal Exemptions from this file since it won't be used anymore. I changed BannedMethods to be included from $SolutionDir instead of anything relative to this package.

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,107 @@
...
+  <!-- By default, the NI.ruleset reports violations as warnings instead of errors. 
+       Set TreatWarningsAsErrors to true in release to generate errors. -->
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <!-- Disable portions of Roslyn analyzers that would otherwise generate errors in our build. There should be a goal
+       around removing as many of these as possible, but we want to report warnings and elevate warnings as errors in
+       our release builds rather than erroring all the time on everything. -->
+  <ItemGroup>
```
jryckman, line 90: This is the content that was in NI.RoslynTransition.targets file that I'm just putting directly in here. I don't see a reason to maintain a separate file, but we could if people feel strongly about it.

DelpireNI: I think I could go either way here. It doesn't quite seem appropriate in this file, but its so little, it doesn't seem worth worrying about adding here. I think if we expect this to grow significantly, it would be nice to have as a separate file.

DelpireNI (approved):

configuration/NI.Analysis.targets
jryckman: Yeah I agree. This is the largest it's every going to be, and it may shrink as time goes on, though I'm not sure by how much

configuration/NI.Analysis.targets
pakdev: Do we expect clients to define `PkgNI_DotNet_Analyzers`? If so, I'd just get rid of "content"

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,110 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <!-- TODO: This location of 'content' needs to get updated once we build our package if we decide to put the files elsewhere, outside of a 'content' directory -->
+    <CodeAnalysisRuleSetDirectory>$(PkgNI_DotNet_Analyzers)\content</CodeAnalysisRuleSetDirectory>
+    <NuGetPackagesDirectory>$(NuGetPackageRoot)</NuGetPackagesDirectory>
+  </PropertyGroup>
+
+	<PropertyGroup>
```
pakdev: Any reason why this is extra indented? Seems like the children aren't 2 spaces either

jryckman: no, I'm not sure why. I'll clean that up and push an update

alejandro5042: Mix of tabs and spaces in this file :)

Replace with spaces, 2 or 4, preferably 4.

docs/Configuring Banned Methods.md
```diff
@@ -0,0 +1,46 @@
+# Configuring Banned Methods (NI1006)
+
+NI1006 is an analyzer rule that prevents specific methods from being used in source. This ensures developers don't accidentally introduce hard-to-find bugs, inappropriate diagnostics, obsolete API usages, etc.
+
+# Process
+
+The analyzer works by looking through "BannedMethods" [additional files](https://github.com/dotnet/roslyn/blob/master/docs/analyzers/Using%20Additional%20Files.md). A project can contain any number of files as long as:
+
```
alejandro5042, line 8: Double blank line, remove?

rulesets/NI.TestUtilities.ruleset
```diff
@@ -0,0 +1,156 @@
...
+    <Rule Id="CA1305" Action="None" />
+    <Rule Id="CA1308" Action="None" />
+    <Rule Id="CA1405" Action="None" />
+    <Rule Id="CA1702" Action="None" />
+    <Rule Id="CA1704" Action="None" />
+    <Rule Id="CA1707" Action="None" />
+    <Rule Id="CA1709" Action="None" />
+    <Rule Id="CA1710" Action="None" />
+    <Rule Id="CA1711" Action="None" />
+    <Rule Id="CA1712" Action="None" /> <!-- RoslynTransition: Should be "Warning". Tracked by Dev Efficiency: https://trello.com/b/gtoe4Of2/iteration-board -->
```
alejandro5042: Should we track this a different way before posting this on GitHub? Who should be tracking this?

rulesets/NI.Tests.ruleset
```diff
@@ -0,0 +1,156 @@
...
+    <Rule Id="CA1305" Action="None" />
+    <Rule Id="CA1308" Action="None" />
+    <Rule Id="CA1405" Action="None" />
+    <Rule Id="CA1702" Action="None" />
+    <Rule Id="CA1704" Action="None" />
+    <Rule Id="CA1707" Action="None" />
+    <Rule Id="CA1709" Action="None" />
+    <Rule Id="CA1710" Action="None" />
+    <Rule Id="CA1711" Action="None" />
+    <Rule Id="CA1712" Action="None" /> <!-- RoslynTransition: Should be "Warning". Tracked by Dev Efficiency: https://trello.com/b/gtoe4Of2/iteration-board -->
```
alejandro5042: Should we track this a different way before posting this on GitHub? Who should be tracking this?

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,110 @@
...
+		<!-- this would more correctly be: assemblyname contains .Test. We have an issue with test utilities projects, and test.testutilities projects. -->
```
alejandro5042: We should document the behavior in these properties somewhere in the README. That way people who use these rules know when some rules apply and others don't.

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,110 @@
...
+		<IsNITestProject Condition="'$(IsNITestProject)' != 'True'">$(MSBuildProjectName.StartsWith("Tests."))</IsNITestProject>
```
alejandro5042: Minor: Consider renaming `IsNITestProject` to `NI_IsTestProject`... cause non-NI people could theoretically use this style guide and NI doesn't claim those projects :)

alejandro5042: Ideally, properties that these `.targets`/`.props` files introduce should be prefixed with something like `NI_` or `NI_CSharpStyleGuide_` so that they don't collide with the user's properties.

alejandro5042 (approved):

configuration/NI.Analysis.targets
jryckman: Yes, I agree that we need to add more documentation to the README. I'll make an action item to do that in a followup submission. I already have a work item for someone to help with the README in general

docs/Configuring Banned Methods.md
jryckman: If I didn't put in two blank lines, then I couldn't get the extra newline I wanted in the preview of the MD.

rulesets/NI.TestUtilities.ruleset
jryckman: We should actually probably just remove all of that.  At this point I think it's stale, and most of these are

configuration/NI.Analysis.targets
pakdev: (Optional) Perhaps we should put a condition on this e.g. `<PropertyGroup Condition="'$(IsNITestProject)' == ''">`. That way, if a project is being classified incorrectly, it can fix itself by manually setting `<IsNITestProject>true|false</IsNITestProject>`. Then, I'd add another PropertyGroup for `IsNITestUtilitiesProject` and one for the ruleset properties (or move them into another).

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,110 @@
...
+		<ProjectNameContainsTests>$(MSBuildProjectName.Contains("Tests"))</ProjectNameContainsTests>
+
+		<CodeAnalysisRuleSet>$(CodeAnalysisRuleSetDirectory)\NI.ruleset</CodeAnalysisRuleSet>
+		<CodeAnalysisRuleSet Condition="'$(IsNITestProjectCondition)' == 'True'">$(CodeAnalysisRuleSetDirectory)\NI.Tests.ruleset</CodeAnalysisRuleSet>
+		<CodeAnalysisRuleSet Condition="'$(IsNITestUtilitiesProjectCondition)' == 'True'">$(CodeAnalysisRuleSetDirectory)\NI.TestUtilities.ruleset</CodeAnalysisRuleSet>
+	</PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Enable Roslyn analyzers if they haven't been explicitly disabled, we're using Roslyn for CA/SA, the code needs to be validated -->
+    <EnableRoslynAnalyzers>True</EnableRoslynAnalyzers>
```
pakdev: This will always enable them. You need to add a condition like `<EnableRoslynAnalyzers Condition="'$(EnableRoslynAnalyzers)' == ''">true</EnableRoslynAnalyzers>`

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,110 @@
...
+      <Link>$(NI1006_BannedMethods)</Link>
+    </AdditionalFiles>
+    <AdditionalFiles Include="$(NI1704_AdditionalSpellingDictionary)" Condition="Exists('$(NI1704_AdditionalSpellingDictionary)')">
+      <Link>$(CodeAnalysisRuleSet)</Link>
+    </AdditionalFiles>
+    <AdditionalFiles Include="$(CodeAnalysisRuleSet)" Condition="Exists('$(CodeAnalysisRuleSet)')">
+     <Link>$(CodeAnalysisRuleSet)</Link>
+    </AdditionalFiles>
+
+    <AdditionalFiles Include="$(NI1704_GlobalExceptions)" Condition="Exists('$(NI1704_GlobalExceptions)')" />
```
pakdev: Perhaps it's too big, but just confirming you didn't want a `<Link />` here too

configuration/NI.Analysis.targets
jryckman: Replaced with 2 spaces which matches our ASW conventions for XML config files.

configuration/NI.Analysis.targets
```diff
@@ -0,0 +1,110 @@
...
+    <RoslynTransition_DisabledRule Include="1712" />
+    <RoslynTransition_DisabledRule Include="1998" />
+  </ItemGroup>
+  <PropertyGroup>
+    <TemporaryWarningsNotAsErrors Condition="'$(TemporaryWarningsNotAsErrors)' == ''"></TemporaryWarningsNotAsErrors>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors),@(RoslynTransition_DisabledRule),$(TemporaryWarningsNotAsErrors)</WarningsNotAsErrors>
+    <NoWarn>$(NoWarn),@(RoslynTransition_DisabledRule)</NoWarn>
+  </PropertyGroup>
+  <ItemGroup>
+    <HashFiles Include="$(MSBuildThisFileFullPath)" />
```
pakdev: ASW uses this for cache invalidation. Rather than assume all projects do, I think we should remove this and let ASW add e.g. `.nuget\packages.config` to HashFiles

configuration/NI.Analysis.targets
jryckman: I have a task in my backlog to add documentation on the different rulesets and which to use when and what behaviors are different.

docs/Configuring Banned Methods.md
```diff
@@ -0,0 +1,46 @@
+# Configuring Banned Methods (NI1006)
+
+NI1006 is an analyzer rule that prevents specific methods from being used in source. This ensures developers don't accidentally introduce hard-to-find bugs, inappropriate diagnostics, obsolete API usages, etc.
+
+# Process
+
+The analyzer works by looking through "BannedMethods" [additional files](https://github.com/dotnet/roslyn/blob/master/docs/analyzers/Using%20Additional%20Files.md). A project can contain any number of files as long as:
```
pakdev: "This analyzer works by loading banned methods from [additional files](url). An additional file's contents will be parsed if the following conditions are true:"

docs/Configuring Banned Methods.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+1. The filename contains the text "BannedMethods"
+2. The contents of the file are XML with the appropriate schema
```
pakdev: Get rid of 2 (or you can add a note that an malformed additional file will output a diagnostic)

configuration/NI.Analysis.targets
jryckman: Good idea. I went with NI_, and updated NI_IsTestProject and NI_IsTestUtilitiesProject

rulesets/NI.TestUtilities.ruleset
```diff
@@ -0,0 +1,156 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="NI.Tests.ruleset" Description="Code analysis rule overrides for NI Test projects." ToolsVersion="14.0">
```
pakdev: Name needs to be updated

pakdev (approved):
Consider some of my comments and fix others

configuration/NI.Analysis.targets
jryckman: After talking with Peter, I'm consolidating the three properties into a single one, NI_EnableAnalyzers and I updated everything accordingly

configuration/NI.Analysis.targets
jryckman: I've removed the hash item. We'll need to remember to update that in ASW though at some point

docs/Configuring Banned Methods.md
jryckman: Nice, thanks! Updated.

docs/Configuring Banned Methods.md
jryckman: I'll get rid of it. I talk about the schema below, and I think that's fine

configuration/NI.Analysis.targets
jryckman: I assume we want a link there. I'll add it.

configuration/NI.Analysis.targets
jryckman: Renamed

rulesets/NI.TestUtilities.ruleset
jryckman: Updated

rulesets/NI.Tests.ruleset
jryckman: I am deleting all of these. Almost all of them are stale and we should just track everything as we find it from here on out in github or at least in the coding practices chapter backlog.

configuration/NI.Analysis.targets
jryckman: I like this idea. I'm going to do this in a follow-up submission where we have a good diff on this one. Adding to backlog.

PR #9, opened 2020-10-01
Remove unused properties in NI.Analysis.targets

jryckman:
# Justification
I removed one property which wasn't used at all, NI_ProjectNameContainsTests.
I removed a property, NI_NuGetPackagesDirectory, since it's basically the same thing as NuGetPackageRoot.

# Implementation
Removed those properties and replaced NI_NuGetPackagesDirectory with NuGetPackageRoot.

# Testing
Can't test the package yet, so I'm adding a few others  to this PR.


jryckman:
The fork pattern that I was trying to follow isn't working as all the commits are clogging things up. I will start using branches off of the fork to clean this up moving forward. 

There are also a few files that look to be deleted as a part of this submission. Those files should have been deleted in the last PR, so if they weren't, this is ensuring those happen.

DelpireNI (approved):

Everest2020 (approved):

PR #16, opened 2020-10-05
Cleanup rulesets by removing duplicate, unused or incorrect configurations

jryckman:
# Justification
There were a lot of rules that are stale, lack implementation in roslyn, or were duplicated. I've cleaned up the files by trimming them down to make them smaller and easier to read/diff. The visual studio editor actually adds duplicates, but I confirmed they aren't necessary. I'm removing them to keep this file as small as possible in the near future.

# Implementation
Hand edited the  files to remove the unnecessary rules. 

# Testing
I verified this on a full build of ASW to ensure there were no build errors.

### DIFF
--- a/rulesets/NI.TestUtilities.ruleset
+++ b/rulesets/NI.TestUtilities.ruleset
@@ -9,13 +9,10 @@
     <Rule Id="CA1008" Action="None" />
     <Rule Id="CA1017" Action="None" />
     <Rule Id="CA1018" Action="None" />
-    <Rule Id="CA1019" Action="Warning" />
     <Rule Id="CA1024" Action="None" />
-    <Rule Id="CA1026" Action="None" />
     <Rule Id="CA1027" Action="None" />
     <Rule Id="CA1030" Action="None" />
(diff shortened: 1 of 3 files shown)


rulesets/NI.TestUtilities.ruleset
```diff
@@ -9,13 +9,10 @@
     <Rule Id="CA1008" Action="None" />
     <Rule Id="CA1017" Action="None" />
     <Rule Id="CA1018" Action="None" />
-    <Rule Id="CA1019" Action="Warning" />
```
jryckman, line 12: Things like this where the TestUtilities/Tests rulesets were more restrictive than the base are wrong. I've just deleted them. Either they differed or were duplicated but had the same action.

gregr-ni: Are you sure none of these should be on in the base?

jryckman: They might be, but I am first getting this repo's files in sync with asw and then I will do another scrub on each override in the tests and utilities. That's coming this week.

PR #27, opened 2020-10-13
Update readme on how to use the package

tsharma-ni:
# Justification
We need to add some instructions on how to use the package, so starting it here.

# Implementation
Added details to the readme. Most of the content stolen from the old readme of the package.

# Testing
Checked out the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,32 +41,6 @@ Inside of the content directory of this NuGet package are the NI rulesets. `NI.r
 If your project starts with `Tests.` or `!Tests`, or ends with `.Tests`, the `NI.Tests.ruleset` will be used. 
 If your project starts with `TestUtilities.` or `!TestUtilities.`, the `NI.TestUtilities.ruleset` will be used.
 
-### Spelling Rules
-By default, this package comes with a dictionary of acceptable spellings. 
-In order to add project specific spellings, create a file called `CodeAnalysisDictionary.xml` 
-in the same directory of your solution file.
-
-Below is an example:
-```xml
-<?xml version="1.0" encoding="utf-8"?>
-<Dictionary>
-  <Words>
-    <Recognized>
-        <Word>example</Word>
-    </Recognized>
-    <Compound>
-      <Term CompoundAlternate="ExampleAlternative">examplealternative</Term>
-    </Compound>
-    <DiscreteExceptions>
-    </DiscreteExceptions>
-  </Words>
-  <Acronyms>
-    <CasingExceptions>
-    </CasingExceptions>
-  </Acronyms>
-</Dictionary>
-```
-
 ## Coding Conventions
 
 - [Purpose of coding conventions](#purpose-of-coding-conventions)
--- a/README.md
+++ b/README.md
@@ -14,7 +14,16 @@ Welcome to NI's internal and external C# conventions and linter rules/plugins/to
 
 ## Quickstart
 
-### Introduction
+### Package Source
+Under the `packageSources` node, add a reference to the source of the package in your [`NuGet.Config`](https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior) file.
+
+```xml
+<packageSources>
+  <add key="<todo-tbd>" value="<todo-tbd>" />
+</packageSources>
+```
+
+### Package Reference
 Add a package reference to the package `NI.CSharp.Analyzers`.
 
 ```msbuild
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Under the `packageSources` node, add a reference to the source of the package in
 Add a package reference to the package `NI.CSharp.Analyzers`.
 
 ```msbuild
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,10 +14,44 @@ Welcome to NI's internal and external C# conventions and linter rules/plugins/to
 
 ## Quickstart
 
-Add a package reference to the package TODO_TBD.
+### Introduction
+Add a package reference to the package `NI.CSharp.Analyzers`.
 
 ```msbuild
-<PackageReference Include="" Version="" />
+<PackageReference Include="NI.CSharp.Analyzers" Version="0.1.0" />
```
Everest2020: this should be adjusted if we want to host the package in GitHub.

Everest2020 (approved):

README.md
tsharma-ni: Yeah, you are right. I completely missed the step of "adding a reference to the package source". I will add that and keep the actual source "to-do".

README.md
```diff
@@ -14,10 +14,55 @@ Welcome to NI's internal and external C# conventions and linter rules/plugins/to
...
+### Package Reference
+Add a package reference to the package `NI.CSharp.Analyzers`.
 
 ```msbuild
-<PackageReference Include="" Version="" />
+<PackageReference Include="NI.CSharp.Analyzers" Version="0.1.0" />
+```
+
+**Note**: Currently, this package only works for projects that reference it using
+[`PackageReference`](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files).
```
jryckman, line 35: You should mention that packages.config doesn't work explicitly

README.md
```diff
@@ -14,10 +14,55 @@ Welcome to NI's internal and external C# conventions and linter rules/plugins/to
...
+
+### Rulesets
+Inside of the content directory of this NuGet package are the NI rulesets. `NI.ruleset` will be applied by default. 
+If your project starts with `Tests.` or `!Tests`, or ends with `.Tests`, the `NI.Tests.ruleset` will be used. 
+If your project starts with `TestUtilities.` or `!TestUtilities.`, the `NI.TestUtilities.ruleset` will be used.
+
+### Spelling Rules
+By default, this package comes with a dictionary of acceptable spellings. 
+In order to add project specific spellings, create a file called `CodeAnalysisDictionary.xml` 
+in the same directory of your solution file.
```
jryckman: replace `of` with `as`

README.md
```diff
@@ -14,10 +14,55 @@ Welcome to NI's internal and external C# conventions and linter rules/plugins/to
...
+
+**Note**: Currently, this package only works for projects that reference it using
+[`PackageReference`](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files).
+
+### Rulesets
+Inside of the content directory of this NuGet package are the NI rulesets. `NI.ruleset` will be applied by default. 
+If your project starts with `Tests.` or `!Tests`, or ends with `.Tests`, the `NI.Tests.ruleset` will be used. 
+If your project starts with `TestUtilities.` or `!TestUtilities.`, the `NI.TestUtilities.ruleset` will be used.
+
+### Spelling Rules
```
jryckman: I think this is a useful document to have, however I think it belongs as a separate document in the /docs folder instead of in the README. We could maybe mention a note on how to configure some of the configurable rules here, but we should link to other documentation, like I did with the banned methods.

jryckman (changes requested):
This is looking good. I made a comment on ripping out your documentation around spelling and putting it into a separate document, like banned methods

README.md
tsharma-ni: Done, fixed in the latest update.

README.md
tsharma-ni: Done, fixed in the latest update.

README.md
tsharma-ni: Done, fixed in the latest update.

README.md
```diff
@@ -14,12 +14,33 @@ Welcome to NI's internal and external C# conventions and linter rules/plugins/to
...
+  <add key="<todo-tbd>" value="<todo-tbd>" />
+</packageSources>
+```
+
+### Package Reference
+Add a package reference to the package `NI.CSharp.Analyzers`.
 
 ```msbuild
-<PackageReference Include="" Version="" />
+<PackageReference Include="NI.CSharp.Analyzers" Version="0.1.0" />
```
jryckman: " Version="0.1.0" /> -- we should look to make something generic here that perhaps lists our versions of the package that are available. I don't think this is something we'll want to update every time we publish a new package.

jryckman (approved):

README.md
tsharma-ni: As of now, maybe `Version="<version-of-your-choice>"` or `Version="<latest-version-available>"`?

Or we could do `Version="1.x.x"`? This would mean on every major version update we would have to update this. Thoughts?

jryckman: I think <package-version> would be sufficient

PR #30, opened 2020-10-15
Enforce "Split chains of dotted method invocations" using roslyn

tsharma-ni:
# Justification
Writing an analyzer for - [Splitting long chains of dotted method/property invocations](https://github.com/ni/csharp/blob/main/docs/Coding-Conventions.md#f34-%EF%B8%8F-consider-splitting-long-chains-of-dotted-methodproperty-invocations)

# Implementation
- Added the analyzer.
- Wrote tests for the new analyzer.

# Testing
The tests that I wrote pass successfully.

**Note**: I am writing a fixer for this convention as well. Will post that in a follow-up PR.


src/NationalInstruments.Analyzers/Properties/Resources.resx
```diff
@@ -345,4 +345,16 @@
   <data name="IdentifiersShouldBeSpelledCorrectlyTitle" xml:space="preserve">
     <value>Identifiers should be spelled correctly</value>
   </data>
+  <data name="NI1017_Description" xml:space="preserve">
+    <value>If you invoke a chain of methods and/or properties and that single line becomes very long, split each invocation onto its own line.</value>
+    <comment>An optional longer localizable description of the diagnostic.</comment>
+  </data>
+  <data name="NI1017_Message" xml:space="preserve">
+    <value>Split long chains of dotted method/property invocations. Use well-named local variables if needed.</value>
```
pakdev: Are we going to enforce this or make it "info" severity? Maybe prefix with "Consider"?

src/NationalInstruments.Analyzers/Style/LongChainDottedInvocationAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecutionIf(IsRunningInProduction);
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterSyntaxNodeAction(AnalyzeExpressionStatementSyntax, SyntaxKind.ExpressionStatement);
+            context.RegisterSyntaxNodeAction(AnalyzeEqualsValueClauseSyntax, SyntaxKind.EqualsValueClause);
+            context.RegisterSyntaxNodeAction(AnalyzeArgumentSyntax, SyntaxKind.Argument);
+            context.RegisterSyntaxNodeAction(AnalyzeArrowExpressionClauseSyntax, SyntaxKind.ArrowExpressionClause);
```
pakdev: Since these individual syntax handlers don't actually do different things, you can just pass all the `SyntaxKind`s in the second parameter (it's `params SyntaxKind[]` I think)

src/NationalInstruments.Analyzers/Style/LongChainDottedInvocationAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+                .FirstOrDefault();
+
+            if (invocationExpressionSyntax is null)
+            {
+                // This does not contain any invocations, bail out
+                return;
+            }
+
+            // Find all invocations which involve brackets ()
+            var nonNestedInvocationsWithParenthesis = invocationExpressionSyntax
```
pakdev: How do we know these aren't nested? Perhaps switch from DescendantNodes to ChildNodes?

src/NationalInstruments.Analyzers/Style/LongChainDottedInvocationAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            // There is a none or only a single call
+            if (nonNestedInvocationsWithParenthesis.Count() < 2)
+            {
+                // No more refactoring required
+                return;
+            }
+
+            // Find all end of lines in the expression
+            var endOfLineTrivias = invocationExpressionSyntax
```
pakdev: What happens if we have something like

```
objects.Where(x => {
   if (x > 2)
   {
       return true;
   }
   else
   {
       return false;
   }
}).Where(x => x % 2 == 0).Select(x => $"num {x}").First();
```

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
```diff
@@ -0,0 +1,331 @@
...
+            var distinctSelectedSoftware = selectedSoftwareContents.GroupBy(software => software.AliasName).Select(g => g.First()).Distinct();
+        }
+
+        private interface ISoftwareContent
+        {
+            string AliasName { get; }
+        }
+    }
+}
+", (11, 44)
```
pakdev: Instead of statically coding this in (which is difficult to determine...especially if the code is dynamically created), you can use the special markers <?>, <|...|>, and <|> where the error should occur. Then, instead of `TestFile`, you use `AutoTestFile` and supply it rules. If you don't want to wait for my follow-up comment, search the existing unit tests for `AutoTestFile` to see usage.

pakdev (commented):
You really knocked this out quickly! Good job

src/NationalInstruments.Analyzers/Properties/Resources.resx
tsharma-ni: Actually, we probably want to promote it to an enforcement. I think that was one of the reasons I started writing this analyzer, the last I talked with Jayson.

@jryckman can you confirm?

src/NationalInstruments.Analyzers/Style/LongChainDottedInvocationAnalyzer.cs
tsharma-ni: > How do we know these aren't nested?

The lambda `syntaxNode => !syntaxNode.IsKind(SyntaxKind.Argument)` in `DescendantNodes()` ensures that we do not get any of the nested invocations. The `DescendantNodes()` call does not go inside any parenthesis (`SyntaxKind.Argument`).

> Perhaps switch from DescendantNodes to ChildNodes?

I tried this actually. But, could not get it to work. See the syntax tree - 

image

The `InvocationExpressionSyntax` are nested, so I needed to call `DescendantNodes()`. Do you have any other ideas?

src/NationalInstruments.Analyzers/Properties/Resources.resx
jryckman: It's part of our  conventions, so it's okay to default to warning

jryckman: I would also note to update the string values. There are two things we want to address: length and complexity. Chained dotted methods are more about complexity than length, though length does matter. I would update the descriptions to talk about complexity over length.

In NI1017_Message, I would also change "long chains" since you're flagging issues once you have two dotted methods that have lambdas. I would change it to "Split chains of dotted method invocations". Don't use property, since we said we were going to skip over properties since they can't have lambdas in them. This has more to do with dotted methods that have lambdas in them.

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
```diff
@@ -0,0 +1,331 @@
...
+        {
+            var test = new TestFile(@"
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NationalInstruments.Analyzers.UnitTests
+{
+    class Test
+    {
+        string Foo(ISoftwareContent softwareContent) => softwareContent.Child().AliasName.Length.ToString();
```
jryckman: Why does `softwareContent.Child().AliasName.Length.ToString()` emit a diagnostic? There's no lambdas in them?

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
```diff
@@ -0,0 +1,331 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NationalInstruments.Analyzers.UnitTests
+{
+    class Test
+    {
+        void Foo(IEnumerable<ISoftwareContent> selectedSoftwareContents)
+        {
+            var distinctSelectedSoftware = selectedSoftwareContents.GroupBy(software => software.AliasName).Select(g => g.First()).Distinct();
```
jryckman: What about a test case where you have multiple dotted methods but the lambdas aren't next to each other:

```
GroupBy(software => software.AliasName).Distinct().Select(g => g.First()).Distinct()
```

I would assume the rule should catch this as well?

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
```diff
@@ -0,0 +1,331 @@
+using System.Collections.Generic;
+using NationalInstruments.Analyzers.Style;
+using NationalInstruments.Analyzers.TestUtilities;
+using NationalInstruments.Analyzers.TestUtilities.TestFiles;
+using NationalInstruments.Analyzers.TestUtilities.Verifiers;
+using Xunit;
+
+namespace NationalInstruments.Analyzers.UnitTests
+{
+    public class LongChainDottedInvocationAnalyzerTests : NIDiagnosticAnalyzerTests<LongChainDottedInvocationAnalyzer>
```
jryckman: Do we have a test case that is using dotted properties along with dotted methods? We only want to create a diagnostic if you have multiple dotted methods with lamdbas as opposed to properties or methods without parameters, since it's the lamdbas that are the huge readability issue.

src/NationalInstruments.Analyzers/Style/LongChainDottedInvocationAnalyzer.cs
tsharma-ni: Thanks, I missed it. I will update.

src/NationalInstruments.Analyzers/Style/LongChainDottedInvocationAnalyzer.cs
tsharma-ni: Thanks. I missed this completely. I will update the logic in the analyzer and also add test to test this kind of code.

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
tsharma-ni: Thanks, I did not know about `AutoTestFile`! I will update the tests.

tsharma-ni:
Had an offline chat with Jayson.

This convention [here](https://github.com/ni/csharp/blob/main/docs/Coding-Conventions.md#f34-%EF%B8%8F-consider-splitting-long-chains-of-dotted-methodproperty-invocations) does not mention lambdas anywhere (looks like it needs to be updated); so, I did not consider the lambdas.

I am going to mark this PR as WIP and rework the analyzer logic. Will mark it ready for review once it's ready again.

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
tsharma-ni: [Fixed](https://github.com/ni/csharp/blob/users/tsharma/add-long-dotted-chain-analyzer/tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs#L221) in the latest update.

src/NationalInstruments.Analyzers/Properties/Resources.resx
tsharma-ni: Fixed in the latest update.

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
tsharma-ni: Yes. Test [here](https://github.com/ni/csharp/blob/users/tsharma/add-long-dotted-chain-analyzer/tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs#L331).

tests/NationalInstruments.Analyzers.UnitTests/LongChainDottedInvocationAnalyzerTests.cs
tsharma-ni: Yes. [Here](https://github.com/ni/csharp/blob/users/tsharma/add-long-dotted-chain-analyzer/tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs#L235) and [here](https://github.com/ni/csharp/blob/users/tsharma/add-long-dotted-chain-analyzer/tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs#L331). One emits a diagnostic, other does not.

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using NationalInstruments.Analyzers.Properties;
+using NationalInstruments.Analyzers.Utilities;
+using NationalInstruments.Analyzers.Utilities.Extensions;
+
+namespace NationalInstruments.Analyzers.Style
+{
+    /// <summary>
+    /// Enforces the convention -
+    /// https://github.com/ni/csharp/blob/main/docs/Coding-Conventions.md#f34-%EF%B8%8F-consider-splitting-long-chains-of-dotted-methodproperty-invocations
```
tsharma-ni, line 16: @jryckman are we going to update this rule in the convention page to mention specifically about lambdas? Or is that going to be a different rule? 

If it's going to be a different rule then I will remove the reference from here.

jryckman: we should update the documentation in the convention page to specifically call out lambdas, but also add a note about readability as well

tsharma-ni: Sure, I'll update the convention page in a follow up PR.

jryckman (approved):

jryckman:
@tsharma-ni  There is another PR that will go in very soon that cleans up the ruleset files: https://github.com/ni/csharp-styleguide/pull/35

The NI.ruleset doesn't list your rule yet because it doesn't exist. As a part of this PR, can you please pull down those changes when they go in and add NI1017 to it? The default value should be whatever you set in your analyzer, so probably warning.

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+                .OfType<InvocationExpressionSyntax>()
+                .FirstOrDefault();
+
+            if (invocationExpressionSyntax is null)
+            {
+                // This does not contain any invocations, bail out
+                return;
+            }
+
+            bool IsLambdaExpression(SyntaxNode syntaxNode) => syntaxNode.IsKind(SyntaxKind.SimpleLambdaExpression);
```
pakdev: Local functions should always go at the bottom

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                // Get the parent invocation expression
+                var parentInvocation = argument.Parent;
+
+                // Find the method/delegate call
+                var memberAccessExpression = parentInvocation?.ChildNodes().OfType<MemberAccessExpressionSyntax>().FirstOrDefault();
+
+                if (memberAccessExpression is null)
+                {
+                    // unknown syntax, don't know what to do
```
pakdev, line 90: We could throw an exception here which will bubble up as AD0001 so clients can report that we haven't considered all cases. The alternative would be them seeing chained methods with lambdas on the same line not being reported and tell us, but that seems unlikely

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+                    break;
+                }
+
+                // Get the dot operator which is used to call the method/delegate
+                var dotToken = memberAccessExpression.ChildTokens().FirstOrDefault(token => token.IsKind(SyntaxKind.DotToken));
+
+                // If the dot operator does not have a leading whitespace, report violation
+                if (!dotToken.LeadingTrivia.Any(trivia => trivia.IsKind(SyntaxKind.WhitespaceTrivia)))
```
pakdev, line 98: So we won't allow short chains on the same line e.g. `values.Where(x => x.Num % 2 == 0).Select(x => x.Num);`?

jryckman: That is my understanding of what we want. I'll loop in @gregr-ni here, but we decided that LINQ statements with lambdas make things harder to read, so always split multiples onto separate lines. It keeps the rules easier to understand as well, rather than also thinking about line length, since we actually have no hard limit on length.

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
```diff
@@ -0,0 +1,648 @@
...
+            string AliasName { get; }
+        }
+    }
+}
+");
+            VerifyDiagnostics(test);
+        }
+
+        [Fact]
+        public void NI1017_OneLambda_NoDiagnostic()
```
pakdev: Could make this a theory with lines 265-268 each value so the actual differences are obvious (sub them into a source template). Same with above

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
```diff
@@ -0,0 +1,648 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NationalInstruments.Analyzers.UnitTests
+{
+    class Test
+    {
+        void Foo(ISoftwareContent softwareContent)
+        {
+            <|softwareContent.Children(x => true).Any(s => s.AliasName == ""T"")|>;
```
pakdev: Since you don't actually use the captured text (stuff that occurs between `<|` and `|>`), you can just place `<|>` at the location of the diagnostic (`<|` in this case). Using the captured text means your diagnostic message to the user would include dynamic text.

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
```diff
@@ -0,0 +1,648 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NationalInstruments.Analyzers.UnitTests
+{
+    class Test
+    {
+        void Foo(ISoftwareContent softwareContent)
+        {
+            <|softwareContent         .Children(x => true)          .Any(s => s.AliasName == ""T"")|>;
```
pakdev: How does this yield a diagnostic if you're just checking for leading trivia?

src/NationalInstruments.Analyzers/Properties/Resources.resx
```diff
@@ -345,4 +345,16 @@
   <data name="IdentifiersShouldBeSpelledCorrectlyTitle" xml:space="preserve">
     <value>Identifiers should be spelled correctly</value>
   </data>
+  <data name="NI1017_Description" xml:space="preserve">
+    <value>If you invoke a chain of methods with lambdas and that single line becomes very long, split each invocation onto its own line.</value>
```
pakdev: Maybe we should get rid of the "very long" verbiage given Jayson's comment elsewhere

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
pakdev: Also, why not search for SimpleLambdaExpressions right off the bat?

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
tsharma-ni: Such whitespaces become [`TrailingTrivia`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxtoken.trailingtrivia?view=roslyn-dotnet) of the previous `SyntaxNode`.

image

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
gregr-ni: What Jayson said. Trying to be more lenient makes it much harder to enforce. This is very similar to requiring all parameters being on separate lines. Yes, there are cases where having more than one on a line would be ok, but it is way simpler to be strict.

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: Done.

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
tsharma-ni: Done.

src/NationalInstruments.Analyzers/Properties/Resources.resx
tsharma-ni: Done. I updated the comment to not talk about the length of the line.

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
tsharma-ni: Done.

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: > Local functions should always go at the bottom

Done. We need an analyzer for this! 😆 

> Also, why not search for SimpleLambdaExpressions right off the bat?

Straight up searching for a `SimpleLambdaExpression` would return me all the lambdas within the expression, even the **nested ones**. For e.g. this statement...

```cs
xyz.GroupBy(software => software.AliasName).Select(g => g.First(x => x != null))
```

...would return three lambdas whereas I really want two.

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: I wasn't sure if it was noticed, so calling it out explicitly. This analyzer is allowing the first lambda expression to stay on the same line. So this...

```cs
values.Where(x => x.Num % 2 == 0)
   .Select(x => x.Num);
```

...is **not** going to generate a diagnostic.

jryckman: I think that makes sense

jryckman: We have a fixer. If we really need to change it, we should be able to apply it to everything at once right?

tsharma-ni: > We have a fixer.

We don't have one yet. At-least, not with this PR. Although I am writing one and will post a follow up soon.

> If we really need to change it, we should be able to apply it to everything at once right?

Yes, changing it should be easy.

pakdev (approved):

PR #32, opened 2020-10-20
Jryckman/update rulesetlogic

jryckman:
# Justification
There was a bug in the .targets file where it was always going to set NI.ruleset as the ruleset because it runs after the csproj defines a <CodeAnalysisRuleSet>. 

# Implementation
The update now checks to see if the ruleset is the default and only override it if it is.

# Testing
Hand testing with custom projects to verify it works

### DIFF
--- a/src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
+++ b/src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
@@ -14,10 +14,14 @@
     <!-- The following is needed for misnamed TestUtilities that contain xaml because the generated project for the second compile appends random characters to the end of the project name -->
     <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True' and '$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.Contains(".TestUtilities"))</NI_IsTestUtilitiesProject>
 
-    <CodeAnalysisRuleSet>$(NI_CodeAnalysisRuleSetDirectory)\NI.ruleset</CodeAnalysisRuleSet>
-    <CodeAnalysisRuleSet Condition="'$(NI_IsTestProjectCondition)' == 'True'">$(NI_CodeAnalysisRuleSetDirectory)\NI.Tests.ruleset</CodeAnalysisRuleSet>
-    <CodeAnalysisRuleSet Condition="'$(NI_IsTestUtilitiesProjectCondition)' == 'True'">$(NI_CodeAnalysisRuleSetDirectory)\NI.TestUtilities.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSetDefined>False</CodeAnalysisRuleSetDefined>
+    <CodeAnalysisRuleSetDefined Condition="'$(CodeAnalysisRuleSet)' != 'MinimumRecommendedRules.ruleset'">True</CodeAnalysisRuleSetDefined>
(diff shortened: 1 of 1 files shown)


src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -14,10 +14,14 @@
     <!-- The following is needed for misnamed TestUtilities that contain xaml because the generated project for the second compile appends random characters to the end of the project name -->
     <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True' and '$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.Contains(".TestUtilities"))</NI_IsTestUtilitiesProject>
 
-    <CodeAnalysisRuleSet>$(NI_CodeAnalysisRuleSetDirectory)\NI.ruleset</CodeAnalysisRuleSet>
-    <CodeAnalysisRuleSet Condition="'$(NI_IsTestProjectCondition)' == 'True'">$(NI_CodeAnalysisRuleSetDirectory)\NI.Tests.ruleset</CodeAnalysisRuleSet>
-    <CodeAnalysisRuleSet Condition="'$(NI_IsTestUtilitiesProjectCondition)' == 'True'">$(NI_CodeAnalysisRuleSetDirectory)\NI.TestUtilities.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSetDefined>False</CodeAnalysisRuleSetDefined>
+    <CodeAnalysisRuleSetDefined Condition="'$(CodeAnalysisRuleSet)' != 'MinimumRecommendedRules.ruleset'">True</CodeAnalysisRuleSetDefined>
```
jryckman: MinimumRecommendedRules.ruleset is the default value that you get if you import Microsoft.CSharp.targets which every csproj needs in order to build with msbuild. I originally tried using empty string but it didn't work and this is why.

pakdev: Thoughts on `<Condition="'$(CodeAnalysisRuleSet)' != '' and '$(CodeAnalysisRuleSet)' != 'MinimumRecommendedRules.ruleset'">`?

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -14,10 +14,14 @@
...
   </PropertyGroup>
+  <PropertyGroup Condition="'$(CodeAnalysisRuleSetDefined)' == 'False'">
```
pakdev, line 21: I think we put blank lines between PGs

pakdev (approved):

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
jryckman: My main concern is if they try to blank out <CodeAnalysisRuleSet></CodeAnalysisRuleSet> because they didn't want one, then we'd overwrite it. I'm less worried about that because, well, they're installing this analyzer package so they want the rules. I'm torn, but I could see going with the '' check if you think it'll cover more cases.

Which case do you think is more likely to happen?

pakdev: I assume if you don't have any ruleset defined the severities will come from each individual analyzer, and I doubt this is the intention of setting a blank ruleset.

PR #33, opened 2020-10-21
Enforce "Do not use LINQ query syntax" using Roslyn

Colin-Tompkins:
# Justification
Created a new analyzer to enforce LINQ method syntax instead of LINQ query syntax.

# Implementation
- Created a new analyzer
- Created a fixer for the analyzer

# Testing
Created new unit tests to verify analyzer and fixer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ If your project starts with `TestUtilities.` or `!TestUtilities.`, the `NI.TestU
   - [[L.12] Properties and Methods](#l12-properties-and-methods)
     - [[L.12.1] ❌ **DO NOT** Introduce side effects when calling properties](#l121--do-not-introduce-side-effects-when-calling-properties)
     - [[L.12.2] ❌ **DO NOT** Allocate each time a property is called](#l122--do-not-allocate-each-time-a-property-is-called)
+  - [[L.13] ❌ **DO NOT** Use LINQ query syntax](#l13--do-not-use-linq-query-syntax)
 - [Documentation and Comments](#documentation-and-comments)
   - [[D.1] ✔️ **DO** Document public and protected members of APIs](#d1-️-do-document-public-and-protected-members-of-apis)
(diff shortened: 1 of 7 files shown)


Everest2020:
Please add a new rule or update the related one for the conventions markdown file.

Colin-Tompkins:
> Please add a new rule or update the related one for the conventions markdown file.

Done.

jryckman (approved):

README.md
```diff
@@ -1077,6 +1078,26 @@ public Foo CurrentFoo
 }
 ```
 
+## [L.13] ❌ **DO NOT** Use LINQ query syntax
```
pakdev, line 1081: Do we need any explanation for why?

pakdev (approved):

README.md
jryckman: I was thinking about that and I do think it could have an explanation for consistency/readability. In the end though, I think it's a preference thing because there's no performance hit.

Colin-Tompkins: I'm happy to add an explanation. Do you have a preference of what it should say?

jryckman: Not entirely, but I do think it should talk about consistency within the file and overall readability. I think we should loop @gregr-ni and @ccifra into this discussion

gregr-ni: To me it is about consistency. We don't need multiple ways to do the same thing and that eliminates another syntax that people have to look out for when reading code.

jryckman: I agree with Greg's response here.

Colin-Tompkins: How's this for an explanation?
>Do not use query/SQL syntax for LINQ expressions; use method/object syntax instead.
>This is to promote consistency within our code.

jryckman: I would say something along the lines of "Having only one way to read/write LINQ expressions increases consistency in our code"

PR #34, opened 2020-10-21
Clean up test rulesets

jryckman:
# Justification
Removed temporary, repo-specific overrides from thet rulesets. These rulesets should reflect our ideal coding conventions and not contain repo-specific tech debt. These are significantly trimmed down and shouldn't change much over time.


gregr-ni:
You only put the localization rules in the utility ruleset. Don't they need to be in the test ruleset also?

jryckman:
> 
> 
> You only put the localization rules in the utility ruleset. Don't they need to be in the test ruleset also?

Yes they needed to be. The changes were unsaved in VS Code. I saved and pushed the updates. Good catch, thanks!

gregr-ni (approved):

PR #35, opened 2020-10-23
Update NI.ruleset

jryckman:
# Justification
Consolidate <Rule> together, remove duplicates and cleaned up file to make it easier to read, diff and maintain.

### DIFF
--- a/src/AnalyzerConfiguration/NI.ruleset
+++ b/src/AnalyzerConfiguration/NI.ruleset
@@ -204,7 +204,10 @@
     <Rule Id="NI1005" Action="Warning" />
     <Rule Id="NI1006" Action="Warning" />
     <Rule Id="NI1007" Action="Warning" />
+    <Rule Id="NI1009" Action="Error" />
     <Rule Id="NI1015" Action="Warning" />
+    <Rule Id="NI1016" Action="Warning" />
+    <Rule Id="NI1018" Action="Warning" />
     <Rule Id="NI1704" Action="Warning" />
   </Rules>
   <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
--- a/src/AnalyzerConfiguration/NI.ruleset
+++ b/src/AnalyzerConfiguration/NI.ruleset
@@ -45,6 +45,9 @@
     <Rule Id="CA1063" Action="Warning" />
     <Rule Id="CA1064" Action="Warning" />
     <Rule Id="CA1065" Action="Warning" />
+    <Rule Id="CA1066" Action="None" />
+    <Rule Id="CA1067" Action="None" />
+    <Rule Id="CA1068" Action="None" />
     <Rule Id="CA1200" Action="None" />
     <Rule Id="CA1300" Action="Warning" />
     <Rule Id="CA1301" Action="Warning" />
@@ -65,7 +68,7 @@
     <Rule Id="CA1507" Action="None" />
     <Rule Id="CA1601" Action="Warning" />
     <Rule Id="CA1700" Action="Warning" />
-    <Rule Id="CA1704" Action="Warning" />
+    <Rule Id="CA1704" Action="None" /> <!-- This spelling analyzer is not implemented. Replaced with NI1704.-->
     <Rule Id="CA1707" Action="Warning" />
     <Rule Id="CA1708" Action="Warning" />
     <Rule Id="CA1710" Action="Warning" />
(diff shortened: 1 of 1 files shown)


src/AnalyzerConfiguration/NI.ruleset
```diff
@@ -211,17 +196,19 @@
...
-    <Rule Id="CA2218" Action="Warning" />
+  <Rules AnalyzerId="NI.CSharp.Analyzers" RuleNamespace="NationalInstruments.Analyzers">        
     <Rule Id="LRN001" Action="Warning" />
     <Rule Id="LRT001" Action="Warning" />
     <Rule Id="NI1001" Action="Warning" />
     <Rule Id="NI1004" Action="None" />
     <Rule Id="NI1005" Action="Warning" />
     <Rule Id="NI1006" Action="Warning" />
     <Rule Id="NI1007" Action="Warning" />
+    <Rule Id="NI1009" Action="Error" />
```
pakdev: Any reason why this is Error vs Warning?

src/AnalyzerConfiguration/NI.ruleset
```diff
@@ -174,20 +171,8 @@
...
-    <Rule Id="CA1822" Action="None" />
-    <Rule Id="CA2119" Action="Warning" />
-    <Rule Id="CA2214" Action="Warning" />
-    <Rule Id="CA2219" Action="Warning" />
-  </Rules>
-  <Rules AnalyzerId="Microsoft.QualityGuidelines.CSharp.Analyzers" RuleNamespace="Microsoft.QualityGuidelines.CSharp.Analyzers">
-    <Rule Id="CA2000" Action="None" />
-    <Rule Id="CA2109" Action="None" />
-    <Rule Id="CA2200" Action="Warning" />
+    <Rule Id="RS0015" Action="None" />
```
pakdev: Re-saving the ruleset rearranges these to be first since their number is lower than RS1001. It also corrects RS1004/1009. So, it would be nice if these were ordered correctly.

src/AnalyzerConfiguration/NI.ruleset
```diff
@@ -399,46 +386,6 @@
...
-  <Rules AnalyzerId="System.Runtime.InteropServices.CSharp.Analyzers" RuleNamespace="System.Runtime.InteropServices.CSharp.Analyzers">
-    <Rule Id="RS0015" Action="None" />
-  </Rules>
-  <Rules AnalyzerId="System.Threading.Tasks.Analyzers" RuleNamespace="System.Threading.Tasks.Analyzers">
-    <Rule Id="RS0018" Action="None" />
-  </Rules>
-  <Rules AnalyzerId="Text.CSharp.Analyzers" RuleNamespace="Text.CSharp.Analyzers">
-    <Rule Id="CA1704" Action="Warning" />
-  </Rules>
   <Rules AnalyzerId="Roslynator.CSharp.Analyzers" RuleNamespace="Roslynator.CSharp.Analyzers">
```
pakdev: Re-saving the ruleset puts these analyzers above StyleCop as these start with "R". Once this change and the other are made, editing this file with the official ruleset editor seems possible (it only removes comments).

pakdev (approved):

src/AnalyzerConfiguration/NI.ruleset
jryckman: I'm just reflecting what the default values are in the analyzers. It really should be warning, but that should also be true in our analyzers as that's normally what we follow.

src/AnalyzerConfiguration/NI.ruleset
jryckman: It's hard to understand what you're calling out specifically since we can't highlight sections of code, but I assume you mean SA vs. RS codes, so I will move RS above SA. If that isn't what you were talking about, please clarify

src/AnalyzerConfiguration/NI.ruleset
jryckman: I misread them and thought that 0015 was 1015. I have ordered everything in this section correctly

src/AnalyzerConfiguration/NI.ruleset
jryckman: There's no reason I have to reflect it though, so I'll change it to warning to be consistent

src/AnalyzerConfiguration/NI.ruleset
jryckman: nevermind, I think you're saying Roslynator vs StyleCop in the AnalyzerId ... I'll sort them all

PR #39, opened 2020-10-29
Update "Splitting long chains of dotted method/property invocations" convention in the readme

tsharma-ni:
# Justification
The [Splitting long chains of dotted method/property invocations](https://github.com/ni/csharp-styleguide#f34-%EF%B8%8F-consider-splitting-long-chains-of-dotted-methodproperty-invocations) needed some updating.

# Implementation
- Update the convention in the readme.
- Update the link of the convention in the analyzer.

# Testing
Checked out the preview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ If your project starts with `TestUtilities.` or `!TestUtilities.`, the `NI.TestU
     - [[F.3.1] ❌ **AVOID** Adding excessive complexity in compound statements](#f31--avoid-adding-excessive-complexity-in-compound-statements)
     - [[F.3.2] ✔️ **CONSIDER** Extracting methods from complex nested expressions](#f32-️-consider-extracting-methods-from-complex-nested-expressions)
     - [[F.3.3] ✔️ **CONSIDER** Splitting long parameter lists](#f33-️-consider-splitting-long-parameter-lists)
-    - [[F.3.4] ✔️ **CONSIDER** Splitting long chains of dotted method/property invocations](#f34-️-consider-splitting-long-chains-of-dotted-methodproperty-invocations)
+    - [[F.3.4] ✔️ **DO** Split chains of method invocations with lambda expressions 💻](#f34-️-do-split-chains-of-method-invocations-with-lambda-expressions)
     - [[F.3.5] ✔️ **DO** Apply consistent formatting on split statements](#f35-️-do-apply-consistent-formatting-on-split-statements)
   - [[F.4] Indents and Braces](#f4-indents-and-braces)
     - [[F.4.1] ✔️ **DO** Use 4 space characters for indentation, never tab characters](#f41-️-do-use-4-space-characters-for-indentation-never-tab-characters)
@@ -456,9 +456,9 @@ ConfigureMyObject(
     downstreamDependencies: originalObjectDependencies.Where(x => x.Direction == Direction.Downstream));
 ```
 
-### [F.3.4] ✔️ **CONSIDER** Splitting long chains of dotted method/property invocations
+### [F.3.4] ✔️ **DO** Split chains of method invocations with lambda expressions 💻
 
-If you invoke a chain of methods and/or properties and that single line becomes very long, split each invocation onto its own line. Use well-named local variables if needed.
+If you invoke a chain of methods with lambda expressions, split each invocation (except the first one) onto its own line. Use well-named local variables if needed.
 
 There is no hard limit on line length. Use judgement to keep code easy to read and diff. See [[F.3.1] ❌ **AVOID** Adding excessive complexity in compound statements](#%5Bf.3.1%5D-%3Ax%3A-**avoid**-adding-excessive-complexity-in-compound-statements).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -456,23 +456,22 @@ ConfigureMyObject(
...
+If you invoke a chain of methods with lambda expressions, split each invocation (except the first one) onto its own line. Use well-named local variables if needed.
 
 There is no hard limit on line length. Use judgement to keep code easy to read and diff. See [[F.3.1] ❌ **AVOID** Adding excessive complexity in compound statements](#%5Bf.3.1%5D-%3Ax%3A-**avoid**-adding-excessive-complexity-in-compound-statements).
 
 Indent dotted invocations one level deeper than the root object. Place the dot for each invocation at the beginning of that line. 
 
 ```csharp
 // Bad
-var distinctSelectedSoftware = selectedSoftwareContents.GroupBy(software => software.AliasName).Select(g => g.First()).Distinct();
+var distinctSoftware = softwareContents.GroupBy(software => software.AliasName).Select(g => g.First()).First(x => x == "Distinct");
```
jryckman: Why did you have to change the // Bad to change the Distinct() to a First? Was it just to have a third lambda? I think we can update this example to be a more realistic case. I'm actually curious to see what the recommendation is there for it. I know that this one is going to get a little contentious otherwise

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -12,8 +12,7 @@
 namespace NationalInstruments.Analyzers.Style
 {
     /// <summary>
-    /// Enforces the convention -
-    /// https://github.com/ni/csharp/blob/main/docs/Coding-Conventions.md#f34-%EF%B8%8F-consider-splitting-long-chains-of-dotted-methodproperty-invocations
+    /// Enforces the convention: [F.3.4] - DO Split chains of method invocations with lambda expressions
```
pakdev: Could use `<see href="link">text</see>`

pakdev (approved):

README.md
```diff
@@ -456,23 +456,22 @@ ConfigureMyObject(
     downstreamDependencies: originalObjectDependencies.Where(x => x.Direction == Direction.Downstream));
 ```
 
-### [F.3.4] ✔️ **CONSIDER** Splitting long chains of dotted method/property invocations
+### [F.3.4] ✔️ **DO** Split chains of method invocations with lambda expressions 💻
 
-If you invoke a chain of methods and/or properties and that single line becomes very long, split each invocation onto its own line. Use well-named local variables if needed.
+If you invoke a chain of methods with lambda expressions, split each invocation (except the first one) onto its own line. Use well-named local variables if needed.
```
jryckman, line 461: @gregr-ni what are your thoughts on this update, specifically the split each invocation (except the first one)? That's how it's currently implemented, but we can easily change that. On one hand, I don't like leaving the first on the first line similar to how we treat parameters in methods, but on the other, I don't love putting each on their own line, since it will potentially end up with extremely short lines of code:

```csharp
var disctinctSelectedSoftware = selectedSoftwareContents.GroupBy(x => x.AliasName)
    .Select(x => x.First());
```
vs
```csharp
var disctinctSelectedSoftware = selectedSoftwareContents
    .GroupBy(x => x.AliasName)
    .Select(x => x.First());
```

jryckman: I want to get @ccifra 's thoughts as well

gregr-ni: I think either should be allowed and the fixer can either leave the first on the same line or use a length heuristic. I would not want the fixer to always go to a second line.

jryckman: @gregr-ni do you think we need to change the wording in this PR? Currently it says (except the first one) but I don't know if it's necessary to change that.

gregr-ni: I'm ok with the wording as is.

jryckman (approved):

README.md
tsharma-ni: Based on our discussion, I updated the example to a more realistic case.

jryckman: I don't see your update. Am I missing something? I still see the old example

jryckman: ah it now says outdated! It should refresh soon

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: Done.

README.md
```diff
@@ -58,7 +58,7 @@ If your project starts with `TestUtilities.` or `!TestUtilities.`, the `NI.TestU
     - [[F.3.1] ❌ **AVOID** Adding excessive complexity in compound statements](#f31--avoid-adding-excessive-complexity-in-compound-statements)
     - [[F.3.2] ✔️ **CONSIDER** Extracting methods from complex nested expressions](#f32-️-consider-extracting-methods-from-complex-nested-expressions)
     - [[F.3.3] ✔️ **CONSIDER** Splitting long parameter lists](#f33-️-consider-splitting-long-parameter-lists)
-    - [[F.3.4] ✔️ **CONSIDER** Splitting long chains of dotted method/property invocations](#f34-️-consider-splitting-long-chains-of-dotted-methodproperty-invocations)
+    - [[F.3.4] ✔️ **DO** Split chains of method invocations with lambda expressions 💻](#f34-️-do-split-chains-of-method-invocations-with-lambda-expressions)
```
jryckman: I don't think this hyperlink is correct. Did you test clicking on F.3.4? I believe you need to add an extra hyphen at the end of the hyperlink to correspond to the computer emoji

PR #44, opened 2020-11-19
Staging changes to add approved namespaces analyzer

rkudukoli:
# Justification
This was originally reviewed and [submitted to AzDo](https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/64349).
Submitting the rule disabled until it is ready to be deployed.
More changes will follow.

# Implementation
Migrated changes from AzDo with minor fixes to account for namespace changes.
Original PR Link - [https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/64349](https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/64349).

# Testing
Migrated unit tests from AzDo

### DIFF
--- a/src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
@@ -70,12 +70,12 @@ namespace NationalInstruments.Tools.Analyzers.Namespaces
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(ProductionRule, TestRule, FileReadRule);
 
         /// <inheritdoc/>
-        public override void Initialize(AnalysisContext analysisContext)
+        public override void Initialize(AnalysisContext context)
         {
-            analysisContext.EnableConcurrentExecutionIf(IsRunningInProduction && !InDebugMode);
-            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecutionIf(IsRunningInProduction && !InDebugMode);
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
-            analysisContext.RegisterCompilationStartAction(OnCompilationStart);
+            context.RegisterCompilationStartAction(OnCompilationStart);
         }
 
         /// <summary>
@@ -187,7 +187,7 @@ namespace NationalInstruments.Tools.Analyzers.Namespaces
                     context.ReportDiagnostic(diagnostic);
                 }
             }
-            
+
             bool ApprovalFilesExist()
             {
                 return !string.IsNullOrEmpty(_approvedNamespacesFilePath)
--- a/src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
@@ -102,8 +102,7 @@ namespace NationalInstruments.Tools.Analyzers.Namespaces
         private void OnCompilationStart(CompilationStartAnalysisContext compilationStartContext)
         {
             InitializeApprovedNamespaces();
-            if (!string.IsNullOrEmpty(_approvedNamespacesFilePath)
-                || !string.IsNullOrEmpty(_approvedTestNamespacesFilePath))
+            if (ApprovalFilesExist())
             {
                 compilationStartContext.RegisterSymbolAction(AnalyzeNamespace, SymbolKind.Namespace);
             }
@@ -187,6 +186,12 @@ namespace NationalInstruments.Tools.Analyzers.Namespaces
                     var diagnostic = Diagnostic.Create(rule, nameSyntax.GetLocation(), namespaceName);
                     context.ReportDiagnostic(diagnostic);
                 }
+            }
+            
+            bool ApprovalFilesExist()
+            {
+                return !string.IsNullOrEmpty(_approvedNamespacesFilePath)
+                    || !string.IsNullOrEmpty(_approvedTestNamespacesFilePath);
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
```diff
@@ -0,0 +1,252 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+using NationalInstruments.Analyzers.Properties;
+using NationalInstruments.Analyzers.Utilities;
+using NationalInstruments.Analyzers.Utilities.Extensions;
+
+namespace NationalInstruments.Tools.Analyzers.Namespaces
```
jryckman: Update the namespace. It doesn't match the layout of the rest of the files in the single assembly now. There are currently two main namespaces: NationalInstruments.Analyzers.Correctness and NationalInstruments.Analyzers.Style

We need to figure out if you want a third, but we should match the hierarchy and ensure that the folder structure matches. Currently this analyzer is inside Correctness

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceCodeFixProvider.cs
```diff
@@ -0,0 +1,50 @@
+using System.Collections.Immutable;
+using System.Composition;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using NationalInstruments.Analyzers.Properties;
+
+namespace NationalInstruments.Tools.Analyzers.Namespaces
```
jryckman: Update the namespace to match the analyzer

jryckman (approved):
When you are ready to enable the review, please add a .md page under /docs that documents how to use this analyzer. There are examples in that folder.

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
rkudukoli: Fixed it.

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceCodeFixProvider.cs
rkudukoli: Fixed it.

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
```diff
@@ -0,0 +1,252 @@
...
+
+        private static bool IsTestNamespace(string namespaceName)
+        {
+            return _testNamespacePattern.IsMatch(namespaceName);
+        }
+
+        private void OnCompilationStart(CompilationStartAnalysisContext compilationStartContext)
+        {
+            InitializeApprovedNamespaces();
+            if (ApprovalFilesExist())
```
pakdev: So this analyzer will just silently do nothing if the additional files don't exist? Is that the behavior we want vs. reporting a diagnostic?

rkudukoli: @pakdev When I turn on this rule, ASW will be the only client that will actually have those files. For other clients, that aren't ready for this rule, should they be forced to create a ruleset where they disable this rule? Are other clients doing this already?

jryckman: I'm a little torn on that. On one hand, it is nice to make an assumption that you have at least one file that can be consumed. However, our goal was that people do not have to create a custom .ruleset to override things in general. I would lean more towards that usability workflow and say that we shouldn't complain if we're missing any files.

pakdev: The rules we enable in this package should be what we wanted enabled for _all_ C# projects at NI. If this rule _only_ applies to ASW, ASW should enable it specifically. Having it NOOP is opaque not just to projects that don't care about it but also to ASW that maybe configures something wrong.

jryckman: That is a good point Pete. Ram, is the idea that only ASW would consume this?

rkudukoli: @pakdev @jryckman , It will be used by ASW initially. However, there is nothing in it that prevents others from using it.

jryckman: I think the best approach is to default the rule to off and then have a specific ruleset in ASW that inherits from the NI.ruleset contained here. It can then turn on the rule and add additional files. That way, the default can be off and we still get the the error reporting behavior that Pete mentioned, which does make sense.

rkudukoli: sounds good.

rkudukoli: done.

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
```diff
@@ -0,0 +1,268 @@
...
+        {
+            return _testNamespacePattern.IsMatch(namespaceName);
+        }
+
+        private void OnCompilationStart(CompilationStartAnalysisContext compilationStartContext)
+        {
+            InitializeApprovedNamespaces();
+            compilationStartContext.RegisterSymbolAction(AnalyzeNamespace, SymbolKind.Namespace);
+
+            if (!ApprovalFilesExist())
```
pakdev, line 118: Thanks for making this change, though it'd probably be more helpful to report diagnostics in `else` cases on lines 154/166. Then you can be more granular and explain that a file named "ApprovedNamespaces.txt" or "ApprovedNamespaces.Tests.txt" is missing. But, that's up to you.

pakdev (approved):

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
rkudukoli: Those files will change going forward. I will leave it as it is for now and iterate on it as new changes are applied

PR #54, opened 2021-01-13
Fix #53 Target file was not updated to reflect change from lib folder to analyzers folder

### LINKED ISSUE #53: Target file was not updated to reflect change from lib folder to analyzers folder.
Latest export is not working. I believe it is because our targets file is still looking in the lib directory for the analyzer dlls.

```xml
    <!-- NI's CA analyzers -->
    <Analyzer Include="$(PkgNI_CSharp_Analyzers)\lib\netstandard1.3\NationalInstruments.Analyzers.Utilities.dll"/>
    <Analyzer Include="$(PkgNI_CSharp_Analyzers)\lib\netstandard1.3\NationalInstruments.Analyzers.dll"/>
```

These should be looking in the analyzers directory.

Everest2020:
In the previous commit, forgot to adjust the library folder in NI.CShapr.Analyzers.nuspec.
It is corrected in this PR, and now the analyzers from NI will be directly put under the dotnet\cs folder instead of dotnet\cs\netstandard1.3

### DIFF
--- a/build/NI.CSharp.Analyzers.nuspec
+++ b/build/NI.CSharp.Analyzers.nuspec
@@ -27,7 +27,7 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="..\.binaries\Release\NationalInstruments.Analyzers\**\*.dll" target="analyzers\dotnet\cs" />
+    <file src="..\.binaries\Release\NationalInstruments.Analyzers\netstandard1.3\*.dll" target="analyzers\dotnet\cs" />
     <file src="..\src\AnalyzerConfiguration\NI.CSharp.Analyzers.targets" target="build" />
     <file src="..\src\AnalyzerConfiguration\*.ruleset" target="content" />
     <file src="..\src\AnalyzerConfiguration\NI1704_AdditionalSpellingDictionary.dic" target="content" />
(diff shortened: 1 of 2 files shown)


jryckman (approved):

build/NI.CSharp.Analyzers.nuspec
```diff
@@ -27,7 +27,7 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="..\.binaries\Release\NationalInstruments.Analyzers\**\*.dll" target="analyzers\dotnet\cs" />
+    <file src="..\.binaries\Release\NationalInstruments.Analyzers\netstandard1.3\*.dll" target="analyzers\dotnet\cs" />
```
pakdev, line 30: I don't care this was changed, but was it necessary? Just trying to understand

pakdev (approved):

build/NI.CSharp.Analyzers.nuspec
Everest2020: I just align it to the same structure as the one from Microsoft Analyzer. As mentioned above, the dlls will be directly put under the dotnet\cs folder instead of dotnet\cs\netstandard1.3

PR #65, opened 2021-06-26
Fix for "NI1017 gives false positives involving array initializers"

### LINKED ISSUE #46: NI1017 gives false positives involving array initializers.
The code below reports a failure of rule NI1017 but it shouldn't. Removing the Max at the end avoids the error but it is on its own line so it shouldn't be a problem.

private static int MaxCounts(IEnumerable<int> numbers)
{
    var max = new[]
        {
            numbers.Count(number => number == 1),
            numbers.Count(number => number == 2),
            numbers.Count(number => number == 3)
        }
        .Max();
    return max;
}

tsharma-ni:
# Justification
#46

# Implementation
Array initializers can have individual invocation expressions separated by commas instead of ".". Tweaked the logic of the analyzer to handle array initializers separately - only looking at direct child nodes of an array initializer and handling individual invocation expressions within the array initializer.

# Testing
- Existing tests pass.
- New tests pass.

### DIFF
--- a/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
@@ -60,9 +60,6 @@ namespace NationalInstruments.Analyzers.Style
                 .FirstOrDefault();
 
             AnalyzeInvocationExpression(invocationExpressionSyntax, context.ReportDiagnostic);
-
-            bool IsNotArrayInitializerSyntax(SyntaxNode syntaxNode) =>
-                !syntaxNode.IsKind(SyntaxKind.ArrayInitializerExpression);
         }
 
         private static void AnalyzeArrayInitializer(SyntaxNodeAnalysisContext context)
@@ -136,8 +133,8 @@ namespace NationalInstruments.Analyzers.Style
             bool IsOutOfScopeSyntax(SyntaxNode syntaxNode) => IsNotArgumentSyntax(syntaxNode) && IsNotArrayInitializerSyntax(syntaxNode);
 
             bool IsNotArgumentSyntax(SyntaxNode syntaxNode) => !syntaxNode.IsKind(SyntaxKind.Argument);
-
-            bool IsNotArrayInitializerSyntax(SyntaxNode syntaxNode) => !syntaxNode.IsKind(SyntaxKind.ArrayInitializerExpression);
         }
+
+        private static bool IsNotArrayInitializerSyntax(SyntaxNode syntaxNode) => !syntaxNode.IsKind(SyntaxKind.ArrayInitializerExpression);
     }
 }
--- a/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
@@ -91,7 +91,7 @@ namespace NationalInstruments.Analyzers.Style
 
             // Find all non-nested arguments which contain a lambda expression
             var nonNestedArgumentsWithLambdas = invocationExpressionSyntax
-                .DescendantNodes(IsNonNested)
+                .DescendantNodes(IsOutOfScopeSyntax)
                 .OfType<ArgumentListSyntax>()
                 .Where(argument => argument.DescendantNodes().Any(IsLambdaExpression));
 
@@ -133,7 +133,7 @@ namespace NationalInstruments.Analyzers.Style
 
             bool IsLambdaExpression(SyntaxNode syntaxNode) => syntaxNode.IsKind(SyntaxKind.SimpleLambdaExpression);
 
-            bool IsNonNested(SyntaxNode syntaxNode) => IsNotArgumentSyntax(syntaxNode) && IsNotArrayInitializerSyntax(syntaxNode);
+            bool IsOutOfScopeSyntax(SyntaxNode syntaxNode) => IsNotArgumentSyntax(syntaxNode) && IsNotArrayInitializerSyntax(syntaxNode);
 
             bool IsNotArgumentSyntax(SyntaxNode syntaxNode) => !syntaxNode.IsKind(SyntaxKind.Argument);
 
--- a/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
@@ -36,18 +36,18 @@ namespace NationalInstruments.Analyzers.Style
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
             context.RegisterSyntaxNodeAction(
-                AnalyzeSyntaxContext,
+                AnalyzeExpression,
                 SyntaxKind.ExpressionStatement,
                 SyntaxKind.EqualsValueClause,
                 SyntaxKind.Argument,
                 SyntaxKind.ArrowExpressionClause);
 
             context.RegisterSyntaxNodeAction(
-                AnalyzeArrayInitializerContext,
+                AnalyzeArrayInitializer,
                 SyntaxKind.ArrayInitializerExpression);
         }
 
-        private static void AnalyzeSyntaxContext(SyntaxNodeAnalysisContext context)
+        private static void AnalyzeExpression(SyntaxNodeAnalysisContext context)
         {
             var parentSyntaxNode = context.Node;
 
@@ -65,7 +65,7 @@ namespace NationalInstruments.Analyzers.Style
                 !syntaxNode.IsKind(SyntaxKind.ArrayInitializerExpression);
         }
 
-        private static void AnalyzeArrayInitializerContext(SyntaxNodeAnalysisContext context)
+        private static void AnalyzeArrayInitializer(SyntaxNodeAnalysisContext context)
         {
             var arrayInitializerSyntaxNode = context.Node;
 
--- a/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
@@ -67,10 +67,10 @@ namespace NationalInstruments.Analyzers.Style
 
         private static void AnalyzeArrayInitializer(SyntaxNodeAnalysisContext context)
         {
-            var arrayInitializerSyntaxNode = context.Node;
+            var arrayInitializerSyntax = context.Node;
 
             // Find only direct child invocation expressions
-            var invocationExpressions = arrayInitializerSyntaxNode
+            var invocationExpressions = arrayInitializerSyntax
                 .ChildNodes()
                 .OfType<InvocationExpressionSyntax>();
 
--- a/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
@@ -55,13 +55,14 @@ namespace NationalInstruments.Analyzers.Style
             // a method/delegate call or a property access
             // or a chain of them
             var invocationExpressionSyntax = parentSyntaxNode
-                .DescendantNodes(IsNotArrayCreationSyntax)
+                .DescendantNodes(IsNotArrayInitializerSyntax)
                 .OfType<InvocationExpressionSyntax>()
                 .FirstOrDefault();
 
             AnalyzeSyntaxNode(invocationExpressionSyntax, context.ReportDiagnostic);
 
-            bool IsNotArrayCreationSyntax(SyntaxNode syntaxNode) => !syntaxNode.IsKind(SyntaxKind.ArrayCreationExpression);
+            bool IsNotArrayInitializerSyntax(SyntaxNode syntaxNode) =>
+                !syntaxNode.IsKind(SyntaxKind.ArrayInitializerExpression);
         }
 
         private static void AnalyzeArrayInitializerContext(SyntaxNodeAnalysisContext context)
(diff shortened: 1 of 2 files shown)


Everest2020:
What about the jagged arrays?

tsharma-ni:
> What about the jagged arrays?

@Everest2020 should be the same syntax - `SyntaxKind.ArrayInitializerExpression`. Do you have a test case in mind?

Everest2020:
> > What about the jagged arrays?
> 
> @Everest2020 should be the same syntax - `SyntaxKind.ArrayInitializerExpression`. Do you have a test case in mind?

How about this test case?

```csharp
        private static int Test(IEnumerable<int> numbers)
        {
            var temp = new int[][]
            {
                new [] {
                numbers.Count(number => number == 1),
                numbers.Count(number => number == 2),
                numbers.Count(number => number == 3)
                },

                new [] {
                numbers.Count(number => number == 1),
                numbers.Count(number => number == 2),
                numbers.Count(number => number == 3)},

                new [] {
                numbers.Count(number => number == 1),
                numbers.Count(number => number == 2),
                numbers.Count(number => number == 3)},

            };
            return 42;
        }
```

tsharma-ni:
> > > What about the jagged arrays?
> > 
> > 
> > @Everest2020 should be the same syntax - `SyntaxKind.ArrayInitializerExpression`. Do you have a test case in mind?
> 
> How about this test case?
> 
> ```cs
>         private static int Test(IEnumerable<int> numbers)
>         {
>             var temp = new int[][]
>             {
>                 new [] {
>                 numbers.Count(number => number == 1),
>                 numbers.Count(number => number == 2),
>                 numbers.Count(number => number == 3)
>                 },
> 
>                 new [] {
>                 numbers.Count(number => number == 1),
>                 numbers.Count(number => number == 2),
>                 numbers.Count(number => number == 3)},
> 
>                 new [] {
>                 numbers.Count(number => number == 1),
>                 numbers.Count(number => number == 2),
>                 numbers.Count(number => number == 3)},
> 
>             };
>             return 42;
>         }
> ```

I added a test for this. Thanks @Everest2020!

jryckman (approved):
I am deferring this to @pakdev  because he's more familiar with analyzer code.

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -36,25 +36,52 @@ public override void Initialize(AnalysisContext context)
...
                 SyntaxKind.Argument,
                 SyntaxKind.ArrowExpressionClause);
+
+            context.RegisterSyntaxNodeAction(
+                AnalyzeArrayInitializerContext,
+                SyntaxKind.ArrayInitializerExpression);
         }
 
-        private static void AnalyzeSyntaxNode(SyntaxNodeAnalysisContext context)
+        private static void AnalyzeSyntaxContext(SyntaxNodeAnalysisContext context)
```
pakdev: `AnalyzeSyntaxContext` is generic and really an "action" that accepts a "context" as its parameter. I'd rename it to something like `AnalyzeExpression` or something more fitting for its `SyntaxKind`s. Same with `AnalyzeArrayInitializerContext`

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -36,25 +36,52 @@ public override void Initialize(AnalysisContext context)
...
             // or a chain of them
             var invocationExpressionSyntax = parentSyntaxNode
-                .DescendantNodes()
+                .DescendantNodes(IsNotArrayCreationSyntax)
                 .OfType<InvocationExpressionSyntax>()
                 .FirstOrDefault();
 
+            AnalyzeSyntaxNode(invocationExpressionSyntax, context.ReportDiagnostic);
+
+            bool IsNotArrayCreationSyntax(SyntaxNode syntaxNode) => !syntaxNode.IsKind(SyntaxKind.ArrayCreationExpression);
```
pakdev, line 64: Do we have to do this check for the `SyntaxKind`s we selected?

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -36,25 +36,52 @@ public override void Initialize(AnalysisContext context)
...
+                .OfType<InvocationExpressionSyntax>();
+
+            // Analyze individual invocation expressions
+            foreach (var invocationExpression in invocationExpressions)
+            {
+                AnalyzeSyntaxNode(invocationExpression, context.ReportDiagnostic);
+            }
+        }
+
+        private static void AnalyzeSyntaxNode(InvocationExpressionSyntax invocationExpressionSyntax, Action<Diagnostic> reportDiagnostic)
```
pakdev: Name this something less generic

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -36,25 +36,52 @@ public override void Initialize(AnalysisContext context)
...
                 .FirstOrDefault();
 
+            AnalyzeSyntaxNode(invocationExpressionSyntax, context.ReportDiagnostic);
+
+            bool IsNotArrayCreationSyntax(SyntaxNode syntaxNode) => !syntaxNode.IsKind(SyntaxKind.ArrayCreationExpression);
+        }
+
+        private static void AnalyzeArrayInitializerContext(SyntaxNodeAnalysisContext context)
+        {
+            var arrayInitializerSyntaxNode = context.Node;
```
pakdev: Drop the "node" suffix (the other analyzers don't call them nodes)

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -36,25 +36,52 @@ public override void Initialize(AnalysisContext context)
...
+
+            // Find only direct child invocation expressions
+            var invocationExpressions = arrayInitializerSyntaxNode
```
pakdev: `invocationExpressionSyntaxes`. My convention is to suffix anything that's syntax with "Syntax" so if I get the symbol I can just call it the same thing w/o "Syntax"

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -98,14 +125,18 @@ private static void AnalyzeSyntaxNode(SyntaxNodeAnalysisContext context)
...
                     var diagnostic = Diagnostic.Create(Rule, invocationExpressionSyntax.GetLocation());
-                    context.ReportDiagnostic(diagnostic);
+                    reportDiagnostic(diagnostic);
                     break;
                 }
             }
 
             bool IsLambdaExpression(SyntaxNode syntaxNode) => syntaxNode.IsKind(SyntaxKind.SimpleLambdaExpression);
 
+            bool IsNonNested(SyntaxNode syntaxNode) => IsNotArgumentSyntax(syntaxNode) && IsNotArrayInitializerSyntax(syntaxNode);
```
pakdev: I don't know what "non-nested" means based on this definition. Is there another name more descriptive (ideally positive vs. negative)?

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
```diff
@@ -642,6 +642,289 @@ private interface ISoftwareContent
...
+
+namespace NationalInstruments.Analyzers.UnitTests
+{
+    class Test
+    {
+        private static int MaxCounts(IEnumerable<int> numbers, ISoftwareContent softwareContent)
+        {
+            var max = new[]
+            {
+                <|>softwareContent.Children(x => false).Count(s => s.AliasName == ""U""),
```
pakdev, line 692: Are we sure we don't want one-liners in array initializers?

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
```diff
@@ -642,6 +642,289 @@ private interface ISoftwareContent
...
+        }
+    }
+}
+",
+GetNI1017Rule(),
+GetNI1017Rule());
+
+            VerifyDiagnostics(test);
+        }
+
```
pakdev, line 927: What about a lambda that creates an array with lambas?

pakdev (dismissed):
Pretty minor comments - thanks for fixing this issue!

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: Fixed [here](https://github.com/ni/csharp-styleguide/pull/65/commits/d2079a7f6644f6a39faa54aa6fa294d8552725c9).

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: How about [this](https://github.com/ni/csharp-styleguide/pull/65/commits/6d602757ba6eaca7afa6c713e5275c3aa57a3766)?

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: Fixed [here](https://github.com/ni/csharp-styleguide/pull/65/commits/f0828a0e8908b2cfaac4199c0551a709ecdeb0df).

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: Fixed [here](https://github.com/ni/csharp-styleguide/pull/65/commits/d6f8e929e94e3ba597336ec3a1a1108980eecd1d).

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: How about [this](https://github.com/ni/csharp-styleguide/pull/65/commits/96736965f3e7d9fa4377598ba4740bf24350ef64)?

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
tsharma-ni: I think so. @jryckman?

tests/NationalInstruments.Analyzers.UnitTests/ChainOfMethodsWithLambdasAnalyzerTests.cs
tsharma-ni: Thanks for pointing this out! Added test case [here](https://github.com/ni/csharp-styleguide/pull/65/commits/a8c67bafec7f0a615a8cee3a99a3cf4029e4f072).

tsharma-ni:
@pakdev thanks for the quick review! 

I have tried my best, but naming is hard, so bear with me.😄

tsharma-ni:
Hi @pakdev, any more feedback?

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -36,25 +36,53 @@ public override void Initialize(AnalysisContext context)
...
+        {
+            var arrayInitializerSyntax = context.Node;
+
+            // Find only direct child invocation expressions
+            var invocationExpressionSyntaxes = arrayInitializerSyntax
+                .ChildNodes()
+                .OfType<InvocationExpressionSyntax>();
+
+            // Analyze individual invocation expressions
+            foreach (var invocationExpression in invocationExpressionSyntaxes)
```
pakdev: `invocationExpression` is a `SyntaxNode`, so I'd suffix it with "Syntax"

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
```diff
@@ -36,25 +36,53 @@ public override void Initialize(AnalysisContext context)
...
             // or a chain of them
             var invocationExpressionSyntax = parentSyntaxNode
-                .DescendantNodes()
+                .DescendantNodes(IsNotArrayInitializerSyntax)
                 .OfType<InvocationExpressionSyntax>()
                 .FirstOrDefault();
 
+            AnalyzeInvocationExpression(invocationExpressionSyntax, context.ReportDiagnostic);
+
+            bool IsNotArrayInitializerSyntax(SyntaxNode syntaxNode) =>
```
pakdev: This is duplicated on line 140; make into a static method

pakdev (dismissed):
Just correct those last couple of things and you're good to go!

pakdev:
> Hi @pakdev, any more feedback?

Sorry it took me so long - GitHub was notifying my personal email account vs. my work one. Apparently you can configure email notification per organization, though, so now my responses should be timely.

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: Fixed [here](https://github.com/ni/csharp-styleguide/pull/65/commits/f8436c60098422a579a4049a0f9b835d1bea260a).

src/NationalInstruments.Analyzers/Style/ChainOfMethodsWithLambdasAnalyzer.cs
tsharma-ni: Fixed [here](https://github.com/ni/csharp-styleguide/pull/65/commits/ddf7d01411a02d42c5cd752cf2aecc9fee28ed70).

pakdev (approved):

PR #68, opened 2021-09-13
Improve DiagnosticDescriptor consistency

pakdev:
# Justification
Current analyzers have inconsistent names, title styles, default severities, and categories for their DiagnosticDescriptors.

# Implementation
Ensured DiagnosticDescriptors have:

1. The name `DiagnosticId`
2. Diagnostic ids without suffixes e.g. `_ParseError`
2. Titles without periods
3. A primary default severity of `Warning`
4. Secondary default severities of `Error` (e.g. configuration file cannot be read)
5. A category name that matches their namespace

# Testing
- Added `NI1800_Foo` to ASW's NI.ruleset and ensured no warnings occurred while building Release. This proves that rulesets configuring suffixed rule ids won't cause problems after this update


src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
```diff
@@ -57,22 +57,22 @@ public sealed class ApprovedNamespaceAnalyzer : NIDiagnosticAnalyzer
         /// Rule for failure to read approved namespaces files.
         /// </summary>
         public static DiagnosticDescriptor FileReadRule { get; } = new DiagnosticDescriptor(
-            $"{DiagnosticId}_ReadError",
+            DiagnosticId,
             new LocalizableResourceString(nameof(Resources.NI1800_FileReadErrorTitle), Resources.ResourceManager, typeof(Resources)),
             new LocalizableResourceString(nameof(Resources.NI1800_FileReadErrorMessage), Resources.ResourceManager, typeof(Resources)),
-            Resources.CategoryNamespaces,
-            DiagnosticSeverity.Warning,
+            Category.IO,
```
jryckman: We spoke about using @pakdev and I spoke about using Category.IO and we're going to remove it so we don't have the same diagnostic ID with multiple categories

jryckman (changes requested):

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceAnalyzer.cs
pakdev: Done

jryckman (approved):

PR #67, opened 2021-09-07
Update analyzers and frameworks to latest versions

pakdev:
# Justification
ASW now [supports](https://ni.visualstudio.com/DevCentral/_git/ASW/commit/238d18d6c98bab350a66e989570c7b3aad07f5f6?refName=refs%2Fheads%2Fmain&tab=details) VS2019 16.9+ ([decision](https://niweb.natinst.com/confluence/display/LV/Everything.sln:+Drop+support+for+VS2017+in+favor+of+VS2019)). This [allows](https://docs.microsoft.com/en-us/visualstudio/extensibility/roslyn-version-support?view=vs-2019#:~:text=.NET%20compiler%20platform%20package%20version%20reference%20%20,2017%20version%2015.7%20%2011%20more%20rows%20) us to use the latest versions of Roslyn analyzers.

# Implementation
- Updated non-analyzer frameworks to be NET Core 5 or NET 4.6.1 (required for a Windows namespace)
- Updated analyzer frameworks to be NET Standard 2.0 (framework used when creating an analyzer through VS template)
- Updated analyzer references to be the latest stable version (except for StyleCop, which has not moved out of beta for years)
- Added project-specific PackageReferences

# Testing
- Ensured code builds and tests pass

+semver: breaking

PR #71, opened 2021-09-14
Fix violations that occur with new analyzers

pakdev:
# Justification
Updating to the latest analyzers in #67 revealed some violations that can/should be solved in the current codebase first.

# Implementation
Solved most of the violations that the latest analyzers complained about. Those that remain require newer APIs.

# Testing
Ensured build/tests still pass.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -44,9 +44,6 @@
 
     <PropertyGroup>
         <IncludeCompatabilityAnalyzer Condition=" '$(IncludeCompatabilityAnalyzer)' == '' ">True</IncludeCompatabilityAnalyzer>
-        <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
-        <RunAnalyzersDuringLiveAnalysis>true</RunAnalyzersDuringLiveAnalysis>
-        <RunAnalyzers>true</RunAnalyzers>
     </PropertyGroup>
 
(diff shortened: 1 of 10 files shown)


Directory.Build.props
```diff
@@ -44,21 +44,20 @@
 
     <PropertyGroup>
         <IncludeCompatabilityAnalyzer Condition=" '$(IncludeCompatabilityAnalyzer)' == '' ">True</IncludeCompatabilityAnalyzer>
+        <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
+        <RunAnalyzersDuringLiveAnalysis>true</RunAnalyzersDuringLiveAnalysis>
+        <RunAnalyzers>true</RunAnalyzers>
```
jryckman: Did we have IncludeCompatibilityAnalyzer set to true before? I'm trying to understand this particular change and why you moved it here

jryckman (approved):
I left one main comment about the move in the property group.

Directory.Build.props
pakdev: Yes; I simply moved these 3 properties into the IncludeCompatbilityAnalyzer's `PropertyGroup`, but I could put them into their own if desired for readability

pakdev: I moved them out of the `PropertyGroup` that only evaluates during debug

pakdev: I just killed them. Apparently they all default to `true`

jryckman: Sounds good

PR #73, opened 2022-07-06
Update dependencies in the nuget spec

yaqian256:
# Justification
Continue the work in #67, it updated the build time dependencies but it did not update the dependencies in the nuget spec, which results in the nuget package still depends on the old dependencies.

# Implementation
- updated the dependencies (except Style Cop and Roslynator.Analyzers) in the nuspec to match the versions used in #67 
- updated StyleCop and Roslynator.Analyzers to latest.
- upgraded non-analyzer frameworks (for the tests only) from NET 4.6.1 to 4.8.0 (the latest GitHub windows build machine does not support 4.6.1 anymore)

# Testing
Ensured code builds and tests pass

### DIFF
--- a/build/NI.CSharp.Analyzers.nuspec
+++ b/build/NI.CSharp.Analyzers.nuspec
@@ -15,15 +15,6 @@
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <license type="file">LICENSE.txt</license>
     <dependencies>
-      <group targetFramework="netstandard1.3">
-        <dependency id="Microsoft.CodeAnalysis.FxCopAnalyzers" version="2.9.10" />
-        <dependency id="Microsoft.CodeQuality.Analyzers" version="2.9.10" />
-        <dependency id="Microsoft.NetCore.Analyzers" version="2.9.10" />
-        <dependency id="Microsoft.NetFramework.Analyzers" version="2.9.10" />
-        <dependency id="Microsoft.VisualStudio.Threading.Analyzers" version="16.3.52" />
-        <dependency id="Roslynator.Analyzers" version="2.1.0" />
-        <dependency id="StyleCop.Analyzers" version="1.1.118" />
-      </group>
       <group targetFramework="netstandard2.0">
         <dependency id="Microsoft.CodeAnalysis.CSharp" version="3.11.0" />
         <dependency id="Microsoft.CodeAnalysis.NetAnalyzers" version="5.0.3" />
--- a/build/NI.CSharp.Analyzers.nuspec
+++ b/build/NI.CSharp.Analyzers.nuspec
@@ -6,7 +6,7 @@
     <!--TODO: author and owner may have to change if we go to nuget.org -->
     <authors>NI</authors>
     <owners>NI</owners>
-    <copyright>Copyright 2020 National Instruments Corporation</copyright>
+    <copyright>Copyright 2022 National Instruments Corporation</copyright>
     <description>NI's code analyzers and rulesets for C# projects.</description>
     <summary>NI's code analyzers and rulesets for C# projects.</summary>
     <icon>images\icon.png</icon>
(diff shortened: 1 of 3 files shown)


jryckman:
@nick-beer fyi, please take a look at this and see if this would be okay for the ASW repo

yaqian256:
> @nick-beer fyi, please take a look at this and see if this would be okay for the ASW repo

There's a related PR for the ASW repo - https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/329321

build/NI.CSharp.Analyzers.nuspec
```diff
@@ -15,14 +15,13 @@
...
-        <dependency id="Microsoft.VisualStudio.Threading.Analyzers" version="16.3.52" />
-        <dependency id="Roslynator.Analyzers" version="2.1.0" />
-        <dependency id="StyleCop.Analyzers" version="1.1.118" />
+      <group targetFramework="netstandard2.0">
+        <dependency id="Microsoft.CodeAnalysis.NetAnalyzers" version="5.0.3" />
+        <dependency id="Microsoft.CodeAnalysis.CSharp" version="3.11.0" />
+        <dependency id="Microsoft.CodeAnalysis.Analyzers" version="3.3.2" />
+        <dependency id="Microsoft.VisualStudio.Threading.Analyzers" version="16.10.56" />
+        <dependency id="StyleCop.Analyzers" version="1.2.0-beta.435" />
+        <dependency id="Roslynator.Analyzers" version="4.1.1" />
```
nick-beer, line 24: We didn't add the Roslynator dependency to .NET Framework.  My recollection is that is was slow, hadn't been enforced anywhere, and would take significant effort to enable.  My feeling is that we should just remove it here and all references to it in our own source bases.  If other repos want it, they can add it themselves.

Thoughts?  @jryckman @pakdev

jryckman: The only concern is that a bit of our style guide depends on the rules from roslynator. Without it, we're going to lose things like parameter spacing, splitting, newlines and a bunch of other things. What part of it wasn't enforced?

nick-beer: As far as I can tell, none of it was enforced via roslynator (in ASW at least...).  It's not referenced in [NI.Analysis.targets](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/BuildTools/NI.Analysis.targets&version=GBmain), so I don't believe it's used for the .NET Framework build.  This also validates that thought:

image


I initially had it turned on for the .NET 6 build, but there were so many violations [that I turned it off](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Source/Directory.Build.NetCore.targets&version=GBmain&line=198&lineEnd=198&lineStartColumn=7&lineEndColumn=114&lineStyle=plain&_a=contents) (especially after seeing that it wasn't enabled for .NET Framework).

jryckman: I thought it was automatically sucked in and didn't need the explicit reference in NI.Analysis.targets. @pakdev do you remember if anything changed about this behavior? I know for a fact that it used to work as I used Rolsynator to update all of the ASW codebase to match our coding conventions, so it definitely used to work without explicit reference in NI.Analysis.targets in the pasts.

pakdev: It needs the explicit reference; perhaps you used Roslynator's CLI tool to [fix all violations in a solution](https://github.com/JosefPihrt/Roslynator/blob/main/docs/HowToFixAllDiagnostics.md)?

See [this discussion](https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/329321?_a=files&path=%2FBuildTools%2FNI.Analysis.targets&discussionId=3333450). It looks like we should just get rid of Roslynator. I'm not sure why the only analyzer we used from it is deprecated (couldn't find any rationale in the change)

jryckman: If we can't use it for the reason we pulled it in, then I agree that we should get rid of it. That leaves a gap in enforcement in our conventions, but it looks like that's currently (and has been) the case. If everything else is okay, let's test this change out with the current ASW.

jryckman (approved):

build/NI.CSharp.Analyzers.nuspec
jryckman: @nick-beer @pakdev I went ahead and approved but I'd like to make sure you're both okay with the changes as well.

nick-beer (approved):
Looks good - I'm excited about this!

build/NI.CSharp.Analyzers.nuspec
pakdev: So are we removing Roslynator in a follow-up change? I confirmed that the latest release doesn't even allow enabling the one rule in our ruleset: RCS1029

jryckman: We should remove it in a subsequent change if we're not using it

PR #74, opened 2022-07-11
Update to .NET 6

yaqian256:
# Justification
Update analyzers to latest version to better support .NET 6.

# Implementation
- Updated analyzer references to be the latest stable version
- Removed `Microsoft.DotNet.Analyzers.Compatibility` because it's been replaced with the built-in analyzers of .NET 5 and later versions.
- Fix new warnings in the code
- Add `global.json`

# Testing
Ensured code builds and tests pass


src/NationalInstruments.Analyzers.Utilities/Renamer.cs
```diff
@@ -1,38 +0,0 @@
...
-            CancellationToken cancellationToken = default(CancellationToken))
-        {
-            if (document == null)
-            {
-                throw new ArgumentNullException(nameof(document));
-            }
-
-            Solution solution = document.Project.Solution;
-
-            return Microsoft.CodeAnalysis.Rename.Renamer.RenameSymbolAsync(
```
yaqian256, line 30: NOTE: this RenameSymbolAsync method (with the `OptionSet` parameter) has been obsoleted and caused build failures. I changed the caller in `FieldsCamelCasedWithUnderscoreCodeFixProvider` to call the new `RenameSymbolAsync` method (with `SymbolRenameOptions`) directly.

Deleted this class because Roslynator had also removed it since v1.3.10.

src/NationalInstruments.Analyzers/Correctness/AwaitInReadLockOrTransactionAnalyzer.cs
```diff
@@ -67,7 +67,7 @@ private void AnalyzeUsingStatement(SyntaxNodeAnalysisContext context)
             var usingStatementSyntax = (UsingStatementSyntax)context.Node;
             var containingMethodSyntax = usingStatementSyntax.Ancestors().OfType<BaseMethodDeclarationSyntax>().FirstOrDefault();
             if (containingMethodSyntax != null &&
-                !containingMethodSyntax.Modifiers.Any(syntaxToken => syntaxToken.Kind() == SyntaxKind.AsyncKeyword))
+                !containingMethodSyntax.Modifiers.Any(syntaxToken => syntaxToken.IsKind(SyntaxKind.AsyncKeyword)))
```
yaqian256: NOTE: this is to fix a warning reported by the new csharp analyzer.

Directory.Build.props
```diff
@@ -58,15 +54,11 @@
...
-        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3" />
-        <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.11.0" />
-        <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.2" />
-        <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="16.10.56" />
+        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0" />
+        <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.2.0" />
+        <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3" />
+        <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.2.32" />
         <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" />
-        <PackageReference Condition=" '$(IncludeCompatabilityAnalyzer)' == 'True' " Include="Microsoft.DotNet.Analyzers.Compatibility" Version="0.2.12-alpha">
```
yaqian256, line 66: NOTE: Removed this because it's been replaced with the built-in analyzers of .NET 5 and later versions,  according to https://www.nuget.org/packages/Microsoft.DotNet.Analyzers.Compatibility/0.2.12-alpha

Directory.Build.props
```diff
@@ -58,15 +54,11 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3" />
-        <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.11.0" />
-        <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.2" />
-        <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="16.10.56" />
+        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0" />
```
nick-beer: I think I'd remove this one - per the recommendation in the note [here](https://docs.microsoft.com/en-us/visualstudio/code-quality/install-net-analyzers?view=vs-2022).  It's included and enabled by default in .NET 5+.

nick-beer:
I'd take this opportunity to add a [`global.json` file](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json) to the repo

nick-beer (approved):

jryckman (approved):

jryckman:
Nick has a few comments that still need addressing and I'll put this in once they're addressed.

Directory.Build.props
yaqian256: removed.

yaqian256:
> I'd take this opportunity to add a [`global.json` file](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json) to the repo

added.  I used the same global.json as in the ASW repo.

src/NationalInstruments.Analyzers/Correctness/AwaitInReadLockOrTransactionAnalyzer.cs
```diff
@@ -67,7 +67,7 @@ private void AnalyzeUsingStatement(SyntaxNodeAnalysisContext context)
             var usingStatementSyntax = (UsingStatementSyntax)context.Node;
             var containingMethodSyntax = usingStatementSyntax.Ancestors().OfType<BaseMethodDeclarationSyntax>().FirstOrDefault();
             if (containingMethodSyntax != null &&
```
pakdev: Note: sucks we can't enforce our [own conventions](https://github.com/ni/csharp-styleguide#f35-%EF%B8%8F-do-apply-consistent-formatting-on-split-statements) w/o RCS1029

src/NationalInstruments.Analyzers/Style/FieldsCamelCasedWithUnderscoreCodeFixProvider.cs
```diff
@@ -25,14 +25,14 @@ namespace NationalInstruments.Analyzers.Style
     [Shared]
     public sealed class FieldsCamelCasedWithUnderscoreCodeFixProvider : CodeFixProvider
     {
-        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(FieldsCamelCasedWithUnderscoreAnalyzer.DiagnosticId);
+        public override sealed ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(FieldsCamelCasedWithUnderscoreAnalyzer.DiagnosticId);
```
pakdev: Curious why this change occurred - did some analyzer flag it?

pakdev (approved):

src/NationalInstruments.Analyzers/Correctness/AwaitInReadLockOrTransactionAnalyzer.cs
jryckman: I miss that. Like I said, I am positive we had it enabled in the past!

global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "6.0.301"
```
nick-beer: I'd update this to specify `"rollForward": "patch"`, and change the version to `6.3.00` - similar to [this](https://github.com/dotnet/roslyn-analyzers/blob/58db3ef1529ce44aef5bd6324deab15f7581c27c/global.json#L17).  This will allow us to keep up to date on the latest security patches without needing to constantly modify this file.

src/NationalInstruments.Analyzers/Correctness/AwaitInReadLockOrTransactionAnalyzer.cs
jryckman: I pushed a change to fix the formatting of this

global.json
jryckman: Thanks for the feedback, Nick. @yaqian256 let me know when you push updates to address Nick's feedback and Pete's comments and I'll work on getting this in.

yaqian256: Updated as suggested. Thanks!

src/NationalInstruments.Analyzers/Style/FieldsCamelCasedWithUnderscoreCodeFixProvider.cs
yaqian256: I think it was done by the CodeMaid extension. But looks like "sealed override" is more popular, at least in ASW (see below), so I discard the change.

```
λ git grep "override sealed" *.cs  | wc
     13      94    1634

λ git grep "sealed override" *.cs  | wc
    346    2549   47264
```

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -36,17 +36,12 @@
 
   <ItemGroup Condition="'$(NI_EnableAnalyzers)' == 'True'">
     <!-- Microsoft's SA analyzers -->
-    <Analyzer Include="$(PkgStyleCop_Analyzers)\analyzers\dotnet\cs\StyleCop.Analyzers.dll"/>
-    <Analyzer Include="$(PkgStyleCop_Analyzers)\analyzers\dotnet\cs\StyleCop.Analyzers.CodeFixes.dll"/>
+    <Analyzer Include="$(PkgStyleCop_Analyzers_Unstable)\analyzers\dotnet\cs\StyleCop.Analyzers.dll"/>
+    <Analyzer Include="$(PkgStyleCop_Analyzers_Unstable)\analyzers\dotnet\cs\StyleCop.Analyzers.CodeFixes.dll"/>
 
     <!-- Microsoft's CA analyzers -->
-    <Analyzer Include="$(PkgMicrosoft_CodeQuality_Analyzers)\analyzers\dotnet\cs\Humanizer.dll" />
```
yaqian256, line 43: NOTE: these are the old packages we do not use anymore.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -36,17 +36,12 @@
 
   <ItemGroup Condition="'$(NI_EnableAnalyzers)' == 'True'">
     <!-- Microsoft's SA analyzers -->
-    <Analyzer Include="$(PkgStyleCop_Analyzers)\analyzers\dotnet\cs\StyleCop.Analyzers.dll"/>
-    <Analyzer Include="$(PkgStyleCop_Analyzers)\analyzers\dotnet\cs\StyleCop.Analyzers.CodeFixes.dll"/>
+    <Analyzer Include="$(PkgStyleCop_Analyzers_Unstable)\analyzers\dotnet\cs\StyleCop.Analyzers.dll"/>
```
yaqian256, line 39: NOTE: we're depending on the beta version of StyleCop to get its latest features.  

The beta version of StyleCop.Analyzers (1.2.0-beta.435) is a thin wrapper of StyleCop.Analyzers.Unstable (1.2.0.435) , 
it does not have the `analyzers` folder. We need to get the analyzers from StyleCop.Analyzers.Unstable .

PR #76, opened 2022-10-05
Add new analyzer for Disposable derivations, with 4 tests

dmondrik:
# Justification
For the `Disposable` base class, we want to avoid calling `GC.SuppressFinalize` and we need to ensure it's safe to do so. The way to do that is to prevent derivations from having a finalizer.

See the [conversation on my ASW PR](https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/373013?_a=files&path=%2FSource%2FCore%2FCore%2FCore%2FDisposable.cs&discussionId=3675859).

# Implementation
Implement an analyzer that creates a diagnostic when a class derived from Disposable has a finalizer.

# Testing
`DisposableOmitsFinalizerAnalyzerTests` includes success and failure test cases.

### DIFF
--- a/src/NationalInstruments.Analyzers.Utilities/Extensions/IMethodSymbolExtensions.cs
+++ b/src/NationalInstruments.Analyzers.Utilities/Extensions/IMethodSymbolExtensions.cs
@@ -1,0 +1,50 @@
+using Microsoft.CodeAnalysis;
+
+namespace NationalInstruments.Analyzers.Utilities.Extensions
+{
+    /// <summary>
+    /// Class that contains helpful extensions to <see cref="IMethodSymbol"/>.
+    /// </summary>
+    /// <remarks>
+    /// Microsoft has most of these extensions internally, but they haven't made them public.
+    /// </remarks>
+    public static class IMethodSymbolExtensions
+    {
+        /// <summary>
+        /// Checks if the given method is a Finalizer implementation.
+        /// </summary>
+        /// <param name="method">The method to check</param>
+        public static bool IsFinalizer(this IMethodSymbol method)
+        {
+            if (method.MethodKind == MethodKind.Destructor)
+            {
+                return true; // for C#
+            }
+
+            if (method.Name != WellKnownMemberNames.DestructorName || !method.Parameters.IsEmpty || !method.ReturnsVoid)
+            {
+                return false;
+            }
+
+            IMethodSymbol overridden = method.OverriddenMethod;
+
+            if (method.ContainingType.SpecialType == SpecialType.System_Object)
+            {
+                // This is object.Finalize
+                return true;
+            }
+
+            if (overridden == null)
+            {
+                return false;
+            }
+
+            for (IMethodSymbol o = overridden.OverriddenMethod; o != null; o = o.OverriddenMethod)
+            {
+                overridden = o;
+            }
+
+            return overridden.ContainingType.SpecialType == SpecialType.System_Object; // it is object.Finalize
+        }
+    }
+}
--- a/src/NationalInstruments.Analyzers.Utilities/Extensions/IMethodSymbolExtensions.cs
+++ b/src/NationalInstruments.Analyzers.Utilities/Extensions/IMethodSymbolExtensions.cs
@@ -12,6 +12,7 @@ namespace NationalInstruments.Analyzers.Utilities.Extensions
     {
         /// <summary>
         /// Checks if the given method is a Finalizer implementation.
+        /// https://github.com/dotnet/roslyn-analyzers/blob/846a766f73caa82608db6fee9f2860004298449f/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs#L114
         /// </summary>
         /// <param name="method">The method to check</param>
         public static bool IsFinalizer(this IMethodSymbol method)
--- a/src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
@@ -19,7 +19,7 @@ namespace NationalInstruments.Analyzers.Correctness
             new LocalizableResourceString(nameof(Resources.NI1816X_MessageFormat), Resources.ResourceManager, typeof(Resources)),
             category: "NationalInstruments",
             DiagnosticSeverity.Warning,
-            isEnabledByDefault: true,
+            isEnabledByDefault: false,
             description: new LocalizableResourceString(nameof(Resources.NI1816X_Description), Resources.ResourceManager, typeof(Resources)));
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
--- a/src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
@@ -1,0 +1,51 @@
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NationalInstruments.Analyzers.Properties;
+using NationalInstruments.Analyzers.Utilities;
+using NationalInstruments.Analyzers.Utilities.Extensions;
+
+namespace NationalInstruments.Analyzers.Correctness
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class DisposableOmitsFinalizerAnalyzer : NIDiagnosticAnalyzer
+    {
+        public const string DiagnosticId = "NI1816X";
+
+        public static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticId,
+            new LocalizableResourceString(nameof(Resources.NI1816X_Title), Resources.ResourceManager, typeof(Resources)),
+            new LocalizableResourceString(nameof(Resources.NI1816X_MessageFormat), Resources.ResourceManager, typeof(Resources)),
+            category: "NationalInstruments",
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: new LocalizableResourceString(nameof(Resources.NI1816X_Description), Resources.ResourceManager, typeof(Resources)));
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+            context.RegisterSymbolAction(AnalyzeSymbol, SymbolKind.NamedType);
+        }
+
+        private static void AnalyzeSymbol(SymbolAnalysisContext context)
+        {
+            var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
+
+            if (namedTypeSymbol.IsOrInheritsFromClass("NationalInstruments.Core.Disposable"))
+            {
+                if (namedTypeSymbol.GetMembers()
+                    .OfType<IMethodSymbol>()
+                    .Any(m => m.IsFinalizer()))
+                {
+                    // For all such symbols, produce a diagnostic.
+                    var diagnostic = Diagnostic.Create(Rule, namedTypeSymbol.Locations[0], namedTypeSymbol.Name);
+                    context.ReportDiagnostic(diagnostic);
+                }
+            }
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


jryckman (commented):
I'm good with these, but Pete is the C# analyzer expert, so I'd like to wait for him to review this. He may not check all the time, so send him an email if you don't hear back soon.

jryckman (approved):

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+            context.RegisterSymbolAction(AnalyzeSymbol, SymbolKind.NamedType);
+        }
+
+        private static void AnalyzeSymbol(SymbolAnalysisContext context)
+        {
+            var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
+
+            if (namedTypeSymbol.IsOrInheritsFromClass("NationalInstruments.Core.Disposable"))
```
nick-beer, line 38: I'm not sure it's appropriate for this rule to be in this analyzer?  This analyzer is used from many different repos, and those repos do not have the `Disposable` - it's only in ASW.  Additionally, we're only able to update to the latest version of this analyzer for .NET 6 projects - not for our .NET Framework ASW projects.  So, by putting it here, we're missing a big chunk of the code that we're like to target.  It would make more sense to me for ASW to have its own analyzer for these types of ASW specific rules.

nick-beer: Looking through some of the analyzers we have in the "correctness" category, it seems there's a lot of ASW specific stuff in there - which is perhaps to be expected given the origins of this analyzer. I suppose one more wouldn't hurt... though I would really like to see us break the implicit ASW dependency here sometime.

nick-beer (approved):
I think we should consider how difficult it would be for us to have our own, specific ASW analyzer and pull out some of these rules that are ASW specific.  Until then, I suppose this doesn't hurt - but know you will not get any .NET Framework coverage with these changes in ASW.

dmondrik:
> I'm good with these, but Pete is the C# analyzer expert, so I'd like to wait for him to review this. He may not check all the time, so send him an email if you don't hear back soon.

Pete was helping me learn how to do this yesterday, so he's aware of the PR.

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
dmondrik: Regarding .NET Framework coverage, I built my own first, and added it to `BuildTools\NI.analysis.targets` and verified it with `build.ps1 -c Release`. It found not only the location I intentionally caused to fail, but also an additional location! Once I get a complete build, I'll include the fix for that other location in my existing PR and I'll reset you.

However, that might be a one-time thing unfortunately. I tried copying the Release build of my local build of the anlayzers over the ones brought in via the analyzers Nuget, and it failed to load some dependencies. I'm guessing the NetCore build uses a different mechanism for this?

jryckman: I agree that we should look to break the implicit dependency to ASW at some point. Up until now, we assumed this analyzer was used in ASW and I believe there might be a few other NI-specific analyzers that are making certain assumptions. We originally designed this github project to be a single package that you can import and get all of the functionality for ASW, but we might be at the point where we need to separate things out now that we have more usage outside of ASW.

One possibility is to have two different projects, an ASW-specific one that imports this package. It could then override some settings and add more analyzers. We'd need to look at the cost-benefit of doing that. @nick-beer feel free to ping me and we can talk about that at some point.

pakdev: Splitting ASW/non-ASW analyzers into their own packages yields the following benefits:

- Prevents non-ASW teams from enabling an analyzer that doesn't apply to them
- Documents which analyzers only apply to ASW
- Allows ASW-only analyzers to be enabled by default
- Decreases NuGet package size (really minor)

Feel free to add any that I missed, but I don't see a _huge_ gain. As long as ASW-only analyzers are disabled by default, ASW can opt-in to them by enabling them in their ruleset files. We can do a better job documenting which analyzers are ASW-only by reorganizing this repo e.g. creating `NationalInstruments.ASW.Analyaers.csproj` (come to think of it, a separate assembly would allow the ASW-only analyzers to be enabled by default; other teams don't need to `<Analyzer Include=...` it)

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public const string DiagnosticId = "NI1816X";
+
+        public static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticId,
+            new LocalizableResourceString(nameof(Resources.NI1816X_Title), Resources.ResourceManager, typeof(Resources)),
+            new LocalizableResourceString(nameof(Resources.NI1816X_MessageFormat), Resources.ResourceManager, typeof(Resources)),
+            category: "NationalInstruments",
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
```
pakdev: Probably should make this `false` for the time being

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
jryckman: The only complexity is knowing which analyzers need to be enabled for ASW. We could export a ruleset to make that easier but ASW would need to choose to use that ruleset instead. I think a namespace/folder separation would help organize this project.

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+            description: new LocalizableResourceString(nameof(Resources.NI1816X_Description), Resources.ResourceManager, typeof(Resources)));
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
```
pakdev: We have a Boolean parameter you can use here that changes execution to be single-threaded. This aids in debugging unit tests:

`context.EnableConcurrentExecutionIf(IsRunningInProduction);`

src/NationalInstruments.Analyzers.Utilities/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <remarks>
+    /// Microsoft has most of these extensions internally, but they haven't made them public.
+    /// </remarks>
+    public static class IMethodSymbolExtensions
+    {
+        /// <summary>
+        /// Checks if the given method is a Finalizer implementation.
+        /// </summary>
+        /// <param name="method">The method to check</param>
+        public static bool IsFinalizer(this IMethodSymbol method)
```
pakdev, line 18: Could add a comment that the source is: https://github.com/dotnet/roslyn-analyzers/blob/846a766f73caa82608db6fee9f2860004298449f/src/Utilities/Compiler/Extensions/IMethodSymbolExtensions.cs#L114

tests/NationalInstruments.Analyzers.UnitTests/DisposableOmitsFinalizerAnalyzerTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            public void Dispose()
+            {
+            }
+        }
+        internal class DerivedDisposable : Disposable
+        {   
+            public DerivedDisposable()
+            {
+            }
+            ~DerivedDisposable()
```
pakdev, line 88: Can you add a test for an expression-bodied finalizer e.g.

`~DerivedDisposable() => Dispose();`

tests/NationalInstruments.Analyzers.UnitTests/DisposableOmitsFinalizerAnalyzerTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+            }
+        }
+    }";
+
+            var test = new TestFile(badCode);
+            var expectedFailure = GetResultAt(14, 24, DisposableOmitsFinalizerAnalyzer.Rule, "DerivedDisposable");
+            VerifyDiagnostics(test, expectedFailure);
```
pakdev: This is fine, but in the future, you can do:

````
    var badCode = @"
using System;
namespace NationalInstruments.Core
{
    internal class Disposable : IDisposable
    {
        public Disposable()
        {
        }
        public void Dispose()
        {
        }
    }
    internal class <|DerivedDisposable|> : Disposable
    {   
        public DerivedDisposable()
        {
        }
        ~DerivedDisposable()
        {
        }
    }
}";

    var test = new AutoTestFile(badCode, new Rule(DisposableOmitsFinalizerAnalyzer.Rule));
    VerifyDiagnostics(test);
````

The other symbols supported are:
- `<?>` (identity position and add supplied argument to message))
- `<|>` (identify position)

These are especially useful when parameterizing tests

pakdev (approved):
Good job!

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
dmondrik: Since it applies only to derivations of `NationalInstruments.Core.Disposable` it won't trigger randomly on other code. I'd prefer to keep it enabled if that's OK.

tests/NationalInstruments.Analyzers.UnitTests/DisposableOmitsFinalizerAnalyzerTests.cs
dmondrik: Good call.

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
dmondrik: Since this discussion is for how to split the analyzers later, I'm going to mark it resolved for now, so it doesn't gate this addition from going in.

nick-beer: I think it would make sense to have the ASW analyzer _in_ ASW, rather than github.  It does make sense that the ASW specific analyzer would pull in this analyzer.  In the meantime, I do think it makes sense for ASW specific analyzers to be disabled by default and enabled in the ASW repo.  We could do this when we bump the version of our NuGet reference in ASW.

@dmondrik - For .NET Framework, we pull in the analyzer in `$/.nuget/packages.config`.  For .NET 6, we pull in the analyzer in `$/Source/Directory.Build.NetCore.targets`.  You'll notice that we reference two different versions.  See ASW PR 329321 for discussion around why we weren't able to update the .NET Framework version.

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
dmondrik: Based on the other conversation I'm marking it disabled by default. Someone can show me how to enable it when we pull it into ASW.

src/NationalInstruments.Analyzers/Correctness/DisposableOmitsFinalizerAnalyzer.cs
pakdev: > I think it would make sense to have the ASW analyzer _in_ ASW, rather than github

Yeah. In this case, we'd want NationalInstruments.Analyzers.Utilities to be its own NuGet so we're not duplicating code.

dmondrik: @nick-beer I went ahead and marked it disabled by default, and someone can show me how to enable it when we pull it into ASW. Also, the reasoning at the time of Jonathan's PR that you referenced (for them to be different) was based on us using .NET framework 4.6.2 at the time. Since then, NI has changed to shipping .NET 4.8 instead. So I think we might want to revisit that decision.

nick-beer: @dmondrik - true, however, I do _not_ expect us to update our references to .NET 4.8 - even though we'll be installing .NET 4.8. Updating would be a pretty good effort, and our stated position on this has been that we'd rather spend any/all of that time migrating to modern .NET rather than making updates to our .NET Framework projects.

PR #78, opened 2023-06-07
Fix issue 77 and add tests

pakdev:
# Justification
@davkean identified a performance issue with FieldsCamelCasedWithUnderscoreCodeFixProvider.

# Implementation
@mavasani determined we were matching symbols that **didn't** match the invalid name when in reality we wanted the reverse.

# Testing
Added code fixer tests. Confirmed these failed to complete (in a reasonable time) with the old code and pass instantly with the new.


jryckman (approved):

davkean:
Thanks folks!

pakdev:
@davkean Thank you for reporting! (sorry the fix took so long)

PR #80, opened 2023-09-29
Add new global analyzer config file support

jonathanou:
# Justification
Rulesets have been deprecated by Microsoft since Visual Studio 2019 16.5,  The general recommendation by Microsoft to enable/disable rules is to go through `.editorconfig` and/or `.globalconfig` (i.e. [Global Analyzer Config](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig)).

In addition, if both a ruleset and `.editorconfig`/`globalconfig` specifies the same rule, the behavior is *undefined* per Microsoft's documentation:  

>The precedence rules for conflicting severity entries from a [ruleset file](https://learn.microsoft.com/en-us/visualstudio/code-quality/using-rule-sets-to-group-code-analysis-rules) and an EditorConfig or global AnalyzerConfig file is undefined.

As more projects need to customize rules, they will undoubtedly enable/disable through `.editorconfig`/`.globalconfig`, and we should switch the NI Analyzer rulesets to the new `.editorconfig`/`globalconfig` to prevent undefined behavior.

# Implementation

- Convert NI Analyzer rulesets to equivalent `.globalconfig` files via Visual Studio [as instructed by Microsoft](https://learn.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2022#convert-an-existing-ruleset-file-to-editorconfig-file).
- Add a new `NI.CSharp.Analyzer.props` file to include the new `.globalconfig` files into [`<GlobalAnalyzerConfigFiles>`](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#naming).
- Update the nuspec file to put `props` file and `.globalconfig` files into `build` directory.
- Update `NI.CSharp.Analyzer.targets` file to continue to use rulesets if less than .NET 6 (to maintain backwards compatibility).

# Testing

Manually built the nupkg with the changes and tested it with an existing C# solution to confirm build is correct.  Without the `globalconfig` files, the codebase had over 10,000 warnings that were disabled by the original ruleset.  The warnings largely went away with the new `globalconfig` files, but I was surprised to see there were still 243 warnings left.

However, upon inspection it seems the warnings are all valid, as the original ruleset also did not disable those warnings.  This behavior also explains past code reviews where I noticed certain warnings like not documenting public types not get raised.  With the new `globalconfig`, those warnings all come up correctly;  so it seems rulesets had bugs that the `.globalconfig` files "fixed".

### DIFF
--- a/build/NI.CSharp.Analyzers.nuspec
+++ b/build/NI.CSharp.Analyzers.nuspec
@@ -27,9 +27,9 @@
   </metadata>
   <files>
     <file src="..\.binaries\Release\NationalInstruments.Analyzers\netstandard2.0\*.dll" target="analyzers\dotnet\cs" />
-    <file src="..\src\AnalyzerConfiguration\*.globalconfig" target="build" />
     <file src="..\src\AnalyzerConfiguration\NI.CSharp.Analyzers.props" target="build" />
     <file src="..\src\AnalyzerConfiguration\NI.CSharp.Analyzers.targets" target="build" />
+    <file src="..\src\AnalyzerConfiguration\*.globalconfig" target="content" />
     <file src="..\src\AnalyzerConfiguration\*.ruleset" target="content" />
     <file src="..\src\AnalyzerConfiguration\NI1704_AdditionalSpellingDictionary.dic" target="content" />
     <file src="..\src\AnalyzerConfiguration\stylecop.json" target="content" />
--- a/build/NI.CSharp.Analyzers.nuspec
+++ b/build/NI.CSharp.Analyzers.nuspec
@@ -27,6 +27,8 @@
   </metadata>
   <files>
     <file src="..\.binaries\Release\NationalInstruments.Analyzers\netstandard2.0\*.dll" target="analyzers\dotnet\cs" />
+    <file src="..\src\AnalyzerConfiguration\*.globalconfig" target="build" />
+    <file src="..\src\AnalyzerConfiguration\NI.CSharp.Analyzers.props" target="build" />
     <file src="..\src\AnalyzerConfiguration\NI.CSharp.Analyzers.targets" target="build" />
     <file src="..\src\AnalyzerConfiguration\*.ruleset" target="content" />
     <file src="..\src\AnalyzerConfiguration\NI1704_AdditionalSpellingDictionary.dic" target="content" />
--- a/src/AnalyzerConfiguration/NI.CSharp.Analyzers.TestUtilities.globalconfig
+++ b/src/AnalyzerConfiguration/NI.CSharp.Analyzers.TestUtilities.globalconfig
@@ -1,4 +1,4 @@
-# NOTE: Requires **VS2019 16.3** or later
+# NOTE: Requires **VS2019 16.7** or later
 is_global = true
 
 # set global level higher than NI.CSharp.Analyzers.globalconfig to override rules specified below
--- a/src/AnalyzerConfiguration/NI.CSharp.Analyzers.TestUtilities.globalconfig
+++ b/src/AnalyzerConfiguration/NI.CSharp.Analyzers.TestUtilities.globalconfig
@@ -1,0 +1,17 @@
+# NOTE: Requires **VS2019 16.3** or later
+is_global = true
+
+# set global level higher than NI.CSharp.Analyzers.globalconfig to override rules specified below
+global_level = 1
+
+# NI.TestUtilities.ruleset
+# Description: Code analysis rule overrides for NI Test Utility projects.
+
+# Avoid empty interfaces. Applying the mix-in pattern is an exception to this rule/
+dotnet_diagnostic.CA1040.severity = none
+
+# Specify CultureInfo. Tests shouldn't have to worry about localizable strings.
+dotnet_diagnostic.CA1304.severity = none
+
+# Specify IFormatProvider. Tests shouldn't have to worry about localizable strings.
+dotnet_diagnostic.CA1305.severity = none
--- a/src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
+++ b/src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
@@ -4,16 +4,8 @@
     <NI_CodeAnalysisRuleSetDirectory>$(PkgNI_CSharp_Analyzers)\content</NI_CodeAnalysisRuleSetDirectory>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <!-- this would more correctly be: assemblyname contains .Test. We have an issue with test utilities projects, and test.testutilities projects. -->
-    <NI_IsTestProject Condition="'$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.StartsWith("Tests."))</NI_IsTestProject>
-    <NI_IsTestProject Condition="'$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.EndsWith(".Tests"))</NI_IsTestProject>
-    <NI_IsTestProject Condition="'$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.StartsWith("!Tests."))</NI_IsTestProject>
-    <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True'">$(MSBuildProjectName.StartsWith("TestUtilities."))</NI_IsTestUtilitiesProject>
-    <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True'">$(MSBuildProjectName.StartsWith("!TestUtilities."))</NI_IsTestUtilitiesProject>
-    <!-- The following is needed for misnamed TestUtilities that contain xaml because the generated project for the second compile appends random characters to the end of the project name -->
-    <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True' and '$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.Contains(".TestUtilities"))</NI_IsTestUtilitiesProject>
-
+  <!-- Rulesets are deprecated by Microsoft and we should only enable it if we're using earlier than .NET 6 -->
+  <PropertyGroup Condition="'$(NETCoreSdkVersion)' == '' or $([MSBuild]::VersionLessThan('$(NETCoreSdkVersion)', '6.0'))">
     <CodeAnalysisRuleSetDefined>False</CodeAnalysisRuleSetDefined>
     <CodeAnalysisRuleSetDefined Condition="'$(CodeAnalysisRuleSet)' != '' and '$(CodeAnalysisRuleSet)' != 'MinimumRecommendedRules.ruleset'">True</CodeAnalysisRuleSetDefined>
   </PropertyGroup>
--- a/src/AnalyzerConfiguration/NI.CSharp.Analyzers.Tests.globalconfig
+++ b/src/AnalyzerConfiguration/NI.CSharp.Analyzers.Tests.globalconfig
@@ -1,4 +1,4 @@
-# NOTE: Requires **VS2019 16.3** or later
+# NOTE: Requires **VS2019 16.7** or later
 is_global = true
 
 # set global level higher than NI.CSharp.Analyzers.globalconfig to override rules specified below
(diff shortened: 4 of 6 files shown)


jryckman:
@jonathanou Which project(s) did you test this with? I'd like to make sure it works with the NXG stack (or what's left of it) if you haven't tested that.

@pakdev could you look into this more and give me your thoughts as well?

jonathanou:
> @jonathanou Which project(s) did you test this with? I'd like to make sure it works with the NXG stack (or what's left of it) if you haven't tested that.
> 
> @pakdev could you look into this more and give me your thoughts as well?

I tested with my team's C# solution, which is an ASP.NET Core web app with 22 projects (including tests).  I did not test the NXG stack, I can do that.

jonathanou:
> > @jonathanou Which project(s) did you test this with? I'd like to make sure it works with the NXG stack (or what's left of it) if you haven't tested that.
> > @pakdev could you look into this more and give me your thoughts as well?
> 
> I tested with my team's C# solution, which is an ASP.NET Core web app with 22 projects (including tests). I did not test the NXG stack, I can do that.

@jryckman I tested it with a subset of Everything.NetCore.sln (98 projects built) and I also saw new warnings pop up.  Popular offenders were:

-  CA1805 (https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1805)
-  CA1829 (https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1829)
-  CA1834 (https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1834)
-  CA1846 (https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1846)
-  CA1848 (https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1848)

These rules are not mentioned at all in NI.Ruleset, and with `<IncludeAll Action="Warning" />` I assume that means any rules not mentioned in the ruleset should default to warning?  The converted `NI.CSharp.Analyzers.globalconfig` has this in response to `IncludeAll`:

```ini
# Default severity for analyzer diagnostics - Requires **VS2019 16.5** or later
dotnet_analyzer_diagnostic.severity = warning
```
If this is the intended behavior, it's not clear to me why the rules mentioned above aren't being reported with NI.Ruleset (is it a bug?);  but at least with .globalconfig, it is now correctly reporting them.

pakdev:
> These rules are not mentioned at all in NI.Ruleset, and with `<IncludeAll Action="Warning" />` I assume that means any rules not mentioned in the ruleset should default to warning?...If this is the intended behavior, it's not clear to me why the rules mentioned above aren't being reported with NI.Ruleset (is it a bug?)

It seems that `<IncludeAll Action="Warning" />` [doesn't work properly](https://github.com/dotnet/roslyn/issues/25409). I checked the first couple of rules that you mentioned ([1805](https://github.com/dotnet/roslyn-analyzers/blob/84fb81c27e0554eadf6b12f97eb52c7cd2803c7e/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/QualityGuidelines/DoNotInitializeUnnecessarily.cs#L26), [1829](https://github.com/dotnet/roslyn-analyzers/blob/84fb81c27e0554eadf6b12f97eb52c7cd2803c7e/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Performance/UseCountProperly.cs#L76), and [1834](https://github.com/dotnet/roslyn-analyzers/blob/84fb81c27e0554eadf6b12f97eb52c7cd2803c7e/src/NetAnalyzers/Core/Microsoft.NetCore.Analyzers/Runtime/PreferConstCharOverConstUnitStringAnalyzer.cs#L29)), and their `RunLevel` is either `IdeHidden_BulkConfigurable` or `IdeSuggestion` - both values that yield a `DiagnosticSeverity` of `Hidden` or `Info`, respectively. Thus, it seems like their absence makes sense in light of that issue, and switching to editorconfig will mitigate the problem.

> @pakdev could you look into this more and give me your thoughts as well?

Searching Teams, it seems like my [only reservation](https://teams.microsoft.com/l/message/19:74451c9552e14770bdb7f9c0214a205e@thread.skype/1565113767648?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=3e327e7e-54bb-4696-aa63-2aa3fcb42cf2&parentMessageId=1564074118101&teamName=Coding%20Practices&channelName=Coding%20Conventions&createdTime=1565113767648) with switching to editorconfig in the past was that it was unenforceable. This was true until [VS 2019 16.3](https://github.com/dotnet/roslyn/issues/33558#issuecomment-562257040) (released shortly after my comment in September 2019). Since we should all be at least at that version (likely 2022+), I think this is a great change that we should embrace. Thanks for tackling this, @jonathanou!

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -4,16 +4,8 @@
...
-    <NI_IsTestProject Condition="'$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.EndsWith(".Tests"))</NI_IsTestProject>
-    <NI_IsTestProject Condition="'$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.StartsWith("!Tests."))</NI_IsTestProject>
-    <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True'">$(MSBuildProjectName.StartsWith("TestUtilities."))</NI_IsTestUtilitiesProject>
-    <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True'">$(MSBuildProjectName.StartsWith("!TestUtilities."))</NI_IsTestUtilitiesProject>
-    <!-- The following is needed for misnamed TestUtilities that contain xaml because the generated project for the second compile appends random characters to the end of the project name -->
-    <NI_IsTestUtilitiesProject Condition="'$(NI_IsTestUtilitiesProject)' != 'True' and '$(NI_IsTestProject)' != 'True'">$(MSBuildProjectName.Contains(".TestUtilities"))</NI_IsTestUtilitiesProject>
-
+  <!-- Rulesets are deprecated by Microsoft and we should only enable it if we're using earlier than .NET 6 -->
```
pakdev, line 7: Do you know if this is something we're required to support?

src/AnalyzerConfiguration/NI.CSharp.Analyzers.TestUtilities.globalconfig
```diff
@@ -0,0 +1,17 @@
+# NOTE: Requires **VS2019 16.3** or later
```
pakdev: According to [this issue](https://github.com/dotnet/roslyn/issues/47707), VS2019 16.7 is required for global config

pakdev (approved):

jryckman (approved):
Let's keep iterating on it to make sure it works for all of our projects, but I agree that this is something that we've always wanted but it has been unable to be enforced through editorconfig

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
nick-beer: This is a good question, but I think I would phrase it a little differently.  I wonder if it makes sense for us to freeze our .NET Framework projects on the current version or earlier, and choose to support only .NET 6+ with subsequent changes to this analyzer package.

I'd really like to be taking things out of this package more than putting more and more into it - and maybe if we froze things for .NET Framework we could do that?  For instance, we really should remove our dependencies on the microsoft analyzer packages that are supplied by the dotnet SDK - that's not needed (and causes problems) for modern .NET.

build/NI.CSharp.Analyzers.nuspec
```diff
@@ -27,6 +27,8 @@
   </metadata>
   <files>
     <file src="..\.binaries\Release\NationalInstruments.Analyzers\netstandard2.0\*.dll" target="analyzers\dotnet\cs" />
+    <file src="..\src\AnalyzerConfiguration\*.globalconfig" target="build" />
```
nick-beer: I don't think we should put the .globalconfig files in the `build` folder.  I'd probably put it in `contentFiles` - but putting it next to the other files in `content` would also be ok.

jonathanou: Unfortunately, `PkgNI_CSharp_Analyzer` isn't defined until *after* the props file is imported, so I'm not able to do `$(PkgNI_CSharp_Analyzer)/content` as the `targets` file was able to do;  I opted to put it in the build folder so that I could use `MSBuildThisFileDirectory`.

However, I guess technically I could do `$(MSBuildThisFileDirectory)../content/NI.CSharp.Analyzer.globalconfig`...I can update that.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
jonathanou: > I wonder if it makes sense for us to freeze our .NET Framework projects on the current version or earlier, and choose to support only .NET 6+ with subsequent changes to this analyzer package.

I went in assuming we had to maintain compatibility for the .NET framework build, but if there's the option to stop supporting it I would definitely be in favor.

ASW's Everything.sln is still using 1.2.14 of this analyzer package, not even the current 2.0.11, so I think it may already be effectively frozen for all intents and purposes?  I don't know if there are other .NET framework builds in the company that want to adopt the latest analyzer changes though.

Who can make the call for stopping support?

build/NI.CSharp.Analyzers.nuspec
nick-beer: Yeah - it's not uncommon at all to see that kind of relative pathing in nuget .props/.targets files.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.TestUtilities.globalconfig
jonathanou: Thanks, updated.  Visual Studio converts rule sets to `.editorconfig` files, and the version support comment was added by Visual Studio.  I later manually modified the file to convert it to `globalconfig`;  so I guess the 16.3 was for `.editorconfig`.

build/NI.CSharp.Analyzers.nuspec
jonathanou: Updated to use relative path to content folder.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
jryckman: Good question on support. I don't know. I think someone like @nick-beer or @gregr-ni could help drive that or point us in the right direction.

jryckman:
@nick-beer @pakdev are there any other outstanding issues in this PR that you guys have?

nick-beer:
> @nick-beer @pakdev are there any other outstanding issues in this PR that you guys have?

Nope, nothing from me.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
gregr-ni: I don't have a fundamental problem with saying that new versions don't support .NET framework or making "breaking" changes like dropping ruleset support in favor of configs. The things I'd think about are the ability to provide patches to the version that .NET framework projects are expected to freeze on and the workflow for projects that support both.

If framework projects are expected to freeze, that implies projects that support framework would continue to use rulesets while core projects would switch to configs. That is not exactly convenient for a project that supports both. Both would need to be maintained and stay in sync.

nick-beer: @gregr-ni - unfortunately, I think we're already in the place you describe - in ASW at least.  We have a set of .ruleset files for .NET Framework and a set of .ruleset files for .NET 6+.  Additionally, we haven't brought .NET Framework forward to the latest version of this package, so we already have issues with patches.

My thought is that this is ok - .NET Framework development is very stable at this point.  We haven't made many changes to this package that are needed by the .NET Framework build, and we haven't changed the rulesets in some time. The problem with being frozen, and the task of keeping things in sync hasn't been significant.

One upside to this is that as we're able to improve things for .NET 6+, we also have the opportunity to improve .NET Framework code that's shared with .NET 6+ code.  We'll be able to uptake new Microsoft analyzers with each .NET SDK release - and because the code is shared, both versions have the opportunity to improve.

I do think that if we want to freeze .NET Framework support, we should do that in a separate PR though - so we can focus on the task. I'm ok with letting this go forward as is.

pakdev:
> @nick-beer @pakdev are there any other outstanding issues in this PR that you guys have?

Me neither

pakdev:
> > @nick-beer @pakdev are there any other outstanding issues in this PR that you guys have?
> 
> Me neither

Sorry, didn't mean to close. I thought "Close with comment" meant close the issue - so dumb

PR #84, opened 2023-11-06
Introduce new analysis mode

jonathanou:
# Justification
Microsoft supports analysis modes that preconfigure code analysis rules (e.g. All, Recommended, Minimum, etc).  Projects can set `<AnalysisMode>` to the preconfigured rulesets to enable/disable different warnings.

While we currently have our own rule specifications, we rely on setting `<AnalysisMode>` to `All` and using a global config file to selectively disable certain rules.  As a result, consumers of our nuget cannot specify a different `<AnalysisMode>`, because our global config is currently applied unconditionally and may override certain rules.

To solve this, we should introduce a new analysis mode "NI" that represents the set of rules that we enable/disable.  Our `props/targets` files will look for the NI analysis mode to decide whether to include our override global config.  If the project elects to use their own `<AnalysisMode>`, then the override global config is NOT included and the project gets the right rules configurations.

# Implementation
- Update `props` file to set `AnalysisMode` to `NI` if not already set.
- Update `props` file to include `NI.CSharp.Analyzers.globalconfig` only if `AnalysisMode` is `NI`.
- Update `targets` file to set per-category `AnalysisMode` only if `AnalysisMode` is `NI` to enable the rulesets we want for each category.
- Additional updated the global config file's global level to -98 to be consistent with Microsoft's global config files.

# Testing
Built nuget file and used ASW to test.  Made sure a private unsealed class generated CA 1852 correctly, and verified in binlog viewer that the global config files are correctly excluded when explicitly setting `<AnalysisMode>None</AnalysisMode>`

### DIFF
--- a/src/AnalyzerConfiguration/NI.CSharp.Analyzers.globalconfig
+++ b/src/AnalyzerConfiguration/NI.CSharp.Analyzers.globalconfig
@@ -1,6 +1,10 @@
 # NOTE: Requires **VS2019 16.7** or later
 is_global = true
-global_level = 0
+
+# The .NET global config for different AnalysisMode categories (e.g. Usage, Performance, Design, etc)
+# have global level set at -99, and we want to be a higher level so that we can override the configured
+# warnings.
+global_level = -98
 
 # NI.ruleset
 # Description: Code analysis rules for NI projects.
@@ -147,6 +151,12 @@ dotnet_diagnostic.CA1505.severity = none
 dotnet_diagnostic.CA1506.severity = none
 
 dotnet_diagnostic.CA1507.severity = none
+
+# CA1508: Avoid dead conditional code
+#   While useful, this rule has a buggy implementation and has too many
+#   false positives, particularly when analyzing code inside loops and
+#   try-finally code.
+dotnet_diagnostic.CA1508.severity = none
 
 dotnet_diagnostic.CA1601.severity = warning
 
--- a/src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
+++ b/src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
@@ -10,6 +10,22 @@
     <CodeAnalysisRuleSetDefined Condition="'$(CodeAnalysisRuleSet)' != '' and '$(CodeAnalysisRuleSet)' != 'MinimumRecommendedRules.ruleset'">True</CodeAnalysisRuleSetDefined>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(AnalysisMode)' == 'NI'">
+    <AnalysisModeStyle Condition="'$(AnalysisModeStyle)' == ''">Default</AnalysisModeStyle>
+    <AnalysisModeSecurity Condition="'$(AnalysisModeSecurity)' == ''">Default</AnalysisModeSecurity>
+    <AnalysisModeDesign Condition="'$(AnalysisModeDesign)' == ''">All</AnalysisModeDesign>
+    <AnalysisModeDocumentation Condition="'$(AnalysisModeDocumentation)' == ''">All</AnalysisModeDocumentation>
+    <AnalysisModeGlobalization Condition="'$(AnalysisModeGlobalization)' == ''">All</AnalysisModeGlobalization>
+    <AnalysisModeInteroperability Condition="'$(AnalysisModeInteroperability)' == ''">All</AnalysisModeInteroperability>
+    <AnalysisModeMaintainability Condition="'$(AnalysisModeMaintainability)' == ''">All</AnalysisModeMaintainability>
+    <AnalysisModeNaming Condition="'$(AnalysisModeNaming)' == ''">All</AnalysisModeNaming>
+    <AnalysisModePerformance Condition="'$(AnalysisModePerformance)' == ''">All</AnalysisModePerformance>
+    <AnalysisModeReliability Condition="'$(AnalysisModeReliability)' == ''">All</AnalysisModeReliability>
+    <AnalysisModeUsage Condition="'$(AnalysisModeUsage)' == ''">All</AnalysisModeUsage>
+
+    <EnforceCodeStyleInBuild Condition="'$(EnforceCodeStyleInBuild)' == ''">true</EnforceCodeStyleInBuild>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(CodeAnalysisRuleSetDefined)' == 'False'">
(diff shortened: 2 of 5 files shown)


src/AnalyzerConfiguration/NI.CSharp.Analyzers.globalconfig
```diff
@@ -148,6 +152,12 @@ dotnet_diagnostic.CA1506.severity = none
 
 dotnet_diagnostic.CA1507.severity = none
 
+# CA1508: Avoid dead conditional code
```
jonathanou, line 155: NOTE:  This is disabled in ASW, and I think it should broadly apply to all NI code.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -26,6 +26,22 @@
     <NI1704_AdditionalSpellingDictionary>$(NI_CodeAnalysisRuleSetDirectory)\NI1704_AdditionalSpellingDictionary.dic</NI1704_AdditionalSpellingDictionary>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(NI_EnableAnalyzers)' == 'True' and '$(AnalysisMode)' == 'NI'">
```
jonathanou, line 29: NOTE:  See https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/584099 for more details on why we need to set per-category to All rather than rely on setting `dotnet_analyzer_diagnostic.severity = warning` in our global config file.

jryckman:
I would like @nick-beer to also look at this and get his thoughts.

@jonathanou What would a consumer of this package do to enable the NI ruleset with this change? It would be nice to test this out and see what files are impacted and how to roll out the change. If we don't change all of the current clients, won't these rules be turned off based on your conditional logic?

jonathanou:
> @jonathanou What would a consumer of this package do to enable the NI ruleset with this change? 

Nothing.  I initialize `<AnalysisMode>` to `NI` in the props file (as long as it isn't initialized by someone else first, like `Directory.Build.props`).  So as long as projects don't explicitly request a different analysis mode, the "NI ruleset" will be in effect by default.

> It would be nice to test this out and see what files are impacted and how to roll out the change. If we don't change all of the current clients, won't these rules be turned off based on your conditional logic?

The current "NI ruleset" is being used in ASW already.  I am taking changes I put in for ASW and moving it here because I think they are broadly applicable.  The only difference is the list of warnings to disable:  I didn't move everything over since some of them were disabled due to situations unique to ASW's code base.

jryckman (approved):

jryckman:
Sounds good, thanks for the info.

jryckman:
Waiting to hear from @pakdev

src/AnalyzerConfiguration/NI.CSharp.Analyzers.globalconfig
```diff
@@ -148,6 +152,12 @@ dotnet_diagnostic.CA1506.severity = none
...
+#   While useful, this rule has a buggy implementation and has too many
```
pakdev, line 156: Were issues ever filed for these problems?

nick-beer (approved):
Looks good to me

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -26,6 +26,22 @@
...
+    <AnalysisModeStyle Condition="'$(AnalysisModeStyle)' == ''">Default</AnalysisModeStyle>
```
pakdev, line 30: Why is this category and the one below set to 'Default' vs 'All'?

pakdev:
Why `AnalysisMode` vs `AnalysisLevel`? This verbiage makes it seem like the latter is deprecating the former:

> Starting in .NET 6, you can omit [\<AnalysisMode\>](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode) in favor of a compound value for the \<AnalysisLevel\> property

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
jonathanou: When enabling all security rules in ASW, there are a ton of violations around dll imports not specifying a default search path.  I had tried to add a default search path but tests started to fail.  I did not have time to go through resolving it, so I went with the default rules for now.

The style rules are set to default because they are very noisy and we don't want them all enabled.  It was simpler to use default and then disable certain rules that were still too noisy.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.globalconfig
jonathanou: There are already a few issues on github that were fixed in the latest .NET 8 analyzer, and at the time I was still using the .NET 7 analyzer.  It's possible some of the bugs are fixed, I can try again in ASW to see if it's better now.

jonathanou:
> Why `AnalysisMode` vs `AnalysisLevel`? This verbiage makes it seem like the latter is deprecating the former:
> 
> > Starting in .NET 6, you can omit [<AnalysisMode>](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode) in favor of a compound value for the <AnalysisLevel> property

I didn't view the description as *deprecating* AnalysisMode, just that it's an option to use the compound value for AnalysisLevel as a *shortcut*.

I could see projects wanting to adjust the AnalysisMode of specific categories for projects (e.g. use NI's ruleset but disable all warnings for Performance category), but I'm having a hard time imagining why projects would want to mix analysis levels for different categories—for example use .NET 7 rules for performance category but .NET 8 rules for usage category. 

If the projects did have such a need, they can still set compound values for the standard Microsoft rulesets like `7.0-recommended`.  Or do category-specific analysis levels like `<AnalysisLevelPerformance>7.0-none</AnalysisLevelPerformance>`.  The only limitation right now is they just can't set `7.0-NI`.

While supporting compound analysis level is definitely possible, from a practical standpoint, the code is not trivial.  We'd have to do some string stripping, and then manually set the analysis level and the analysis mode in order to get the right effect.  It's definitely possible, but I feel it's complicated for something that may have little use.

pakdev:
Ah, you're right. It only means you can specify `AnalysisMode` inside `AnalysisLevel` making the former discrete setting irrelevant.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.globalconfig
pakdev: Ok, created #85 to capture this work

pakdev (approved):

PR #88, opened 2023-12-04
Update to use the 8.0.0 analyzer

jonathanou:
# Justification
With .NET 8 released, the Microsoft Net Analyzer also has a new 8.0.0 analyzer available.  We should update our NI analyzer to use the new 8.0.0 analyzer so that consuming projects can check .NET 8 warnings.

# Implementation
- Update the analyzer version to 8.0.0.
- Update the default analysis level to 8.0 to enable .NET 8 warning checking.

# Testing
Built the nuget locally and tested with building ASW.  Verified in the binlog that compilation is now using the 8.0.0 analyzer:

image

image

### DIFF
--- a/build/NI.CSharp.Analyzers.nuspec
+++ b/build/NI.CSharp.Analyzers.nuspec
@@ -16,6 +16,11 @@
     <license type="file">LICENSE.txt</license>
     <dependencies>
       <group targetFramework="netstandard2.0">
+        <!-- 
+          IMPORTANT: when updating Microsoft.CodeAnalysis.NetAnalayzers to a new
+          major version, make sure to update AnalysisLevel in NI.CSharp.Analysers.props
+          too, otherwise new rules will not be properly enabled.
+        -->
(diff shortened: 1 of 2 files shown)


build/NI.CSharp.Analyzers.nuspec
```diff
@@ -16,7 +16,7 @@
     <license type="file">LICENSE.txt</license>
     <dependencies>
       <group targetFramework="netstandard2.0">
-        <dependency id="Microsoft.CodeAnalysis.NetAnalyzers" version="7.0.4" />
+        <dependency id="Microsoft.CodeAnalysis.NetAnalyzers" version="8.0.0" />
```
pakdev, line 24: Maybe we should add a comment here to update `AnalysisLevel` in NI.CSharp.Analyzers.props so they don't get out of sync

pakdev (approved):

build/NI.CSharp.Analyzers.nuspec
jonathanou: Good point, comment added.

jryckman (approved):

PR #94, opened 2024-03-12
Add a new analyzer for checking record equality implementations

jonathanou:
# Justification
For C# `record` types, which are meant to represent a value type, the compiler will automatically generate `Equals` and `GetHashCode` such that the members of the `record` are compared for equality—this contrasts to `class` where the default `Equals` behavior is to compare reference equality of the `class` instance.

However, if the `record` contains members of `IEnumerable`, the default behavior of the generated `Equals` implementation is to compare the *reference equality* of the enumerable member, not the more expected *value* equality of each enumerable element.

Thus for any `record` that contains enumerables, it is important to manually implement `Equals` and `GetHashCode` such that the enumerables are compared via `SequenceEqual` if ordering matters, or `UnorderedEquals`—a helper method provided by NationalInstruments.Core—if not.

In the past, we've relied on code reviews to enforce this requirement, but it can get tedious and easy to miss.  Additionally, it does not allow other teams that are not aware of the default `record` behavior to learn about the requirement.

By providing a custom analyzer, we can auto enforce that all `record` types with enumerables must provide a manually implemented `Equals` and `GetHashCode`, making code reviews easier and educates others about why the default `record` behavior doesn't work for enumerables.

# Implementation
- Add a new analyzer that checks for a `record` type's properties to see if there are enumerables.
- Raise a new `NI1019` warning if the `record` has enumerable properties but doesn't implement `Equals`.

# Testing
Added several unit tests to verify the new warning fires or doesn't fire for various scenarios.

Built the analyzer and tested it in ASW codebase, made sure it shows the warning properly:

image

Encouragingly, the analyzer also found existing `record` violations in ASW that I will work on fixing later.


src/AnalyzerConfiguration/NI.CSharp.Analyzers.globalconfig
```diff
@@ -231,6 +231,17 @@ dotnet_diagnostic.CA1825.severity = none
...
+# CA1852: Seal internal types
+#   CA 1852 is already enabled, but by default, the analyzer treats internal types as public
+#   if the assembly has InternalsVisibleTo attribute defined.  Most of our assemblies have
+#   InternalsVisibleTo defined for at least test projects, and thus checking internal classes
+#   will basically be disabled for our code base.
+#
+#   By setting ignore_internalsvisibleto to true, we ensure internal classes are checked.  For
+#   any unsealed internal class which the test class needs to derive from, we will require developers to
+#   manually suppress CA 1852 and explain why.
+dotnet_code_quality.CA1852.ignore_internalsvisibleto = true
```
jonathanou, line 243: NOTE:  This is unrelated to this PR, but was something I enabled in ASW and thought it would be generally useful.  

I tested this in ASW by deleting the entry in ASW's `.globalconfig` and making sure using the updated analyzer with this change still generated CA1852 for internal types in assemblies that had `InternalsVisibleTo` defined.

pakdev:
Did you intentionally not check for a custom `GetHashCode` implementation? I assume someone already decided that exposing enumerables in records is fine in the first place (it seems like there could be [issues](https://stackoverflow.com/a/69840868) beyond value equality)?

jonathanou:
> Did you intentionally not check for a custom `GetHashCode` implementation?

Yes.  There is already a built-in [C# warning CS0659](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0659) that emits a warning if you implement `Equals` but not `GetHashCode`, so I did not check for it (FYI this is also documented in the code as a comment).

> I assume someone already decided that exposing enumerables in records is fine in the first place

That's definitely possible.  I know some developers like to use record for its concise syntax shortcut (i.e. primary constructor syntax) to quickly create data structures that hold a bunch of variables, e.g:

```cs
internal record MyRecord(int MyInt, string MyString);
```

In those cases, it's true that equality isn't the main concern.  However, starting in .NET 8, [C# 12 supports primary constructor syntax for classes and structs](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/instance-constructors#primary-constructors) too—so using records should now be reserved for truly wanting a value type class where equality works as expected.

And in my experience in code reviews, most are unaware of the enumerable equality problem (or forgot about it).  For my team's code, equality is critically important, and we've had bugs in the past due to failure to override equality.

> it seems like there could be [issues](https://stackoverflow.com/a/69840868) beyond value equality?

The issue you referenced seems to stem from creating a custom enumerable type that performs value equality internally in its `Equals` implementation, so that the default record equality implementation will work out of the box.

That's a different approach than what my team's project has done.  My team's project also uses standard immutable collection types in records heavily for json serialization/deserialization, and we've not run into the serialization issues mentioned in stackoverflow.

src/NationalInstruments.Analyzers/Correctness/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs
```diff
@@ -0,0 +1,95 @@
...
+            Category.Correctness,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: new LocalizableResourceString(nameof(Resources.NI1019_Description), Resources.ResourceManager, typeof(Resources)));
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
```
pakdev, line 32: We typically add the following above line 32:

`context.EnableConcurrentExecutionIf(IsRunningInProduction);`

jonathanou: Updated.

src/NationalInstruments.Analyzers/Correctness/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs
```diff
@@ -0,0 +1,96 @@
...
+                return;
+            }
+
+            var baseTypeProperties = GetBaseTypeProperties(typeSymbol)
+                .Select(p => p.Name)
+                .ToImmutableHashSet();
+
+            var enumerableProperties = typeSymbol
+                .GetPublicPropertySymbols()
+                .Where(p => p.Type.IsEnumerable() && !baseTypeProperties.Contains(p.Name))
```
pakdev: What's the point of ignoring the enumerable if the base type has a property with the same name? Doesn't that allow the following to _not_ report a diagnostic?

```cs
public record BaseRecord
{
    public int MyInts { get; }
}

public record TestRecord : BaseRecord
{
    public IEnumerable<int> MyInts { get; }
}
```

src/NationalInstruments.Analyzers/Correctness/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs
```diff
@@ -0,0 +1,96 @@
...
+                .ToImmutableArray();
+
+            if (enumerableProperties.Length == 0)
+            {
+                return;
+            }
+
+            // check if the record type has implemented its own Equality methods
+            foreach (var location in typeSymbol.Locations)
```
pakdev: Rather than check for the Equals method syntax, what about the following:

```cs
// Convert to INamedTypeSymbol vs. TypedSymbol at start
var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
...
if (namedTypeSymbol.ExplicitlyImplementsEquals())
{
    return;
}
```

Create INamedTypeSymbolExtensions.cs in NationalInstruments.Analyzers.Utilities.Extensions:

```cs
public static class NamedTypeSymbolExtensions
{
    public static bool ExplicitlyImplementsEquals(this INamedTypeSymbol namedTypeSymbol)
        => namedTypeSymbol.GetMembers().OfType<IMethodSymbol>().Any(m => m.IsExplicitEquals());
}
```

Add to IMethodSymbolExtensions.cs:

```cs
public static bool IsExplicitEquals(this IMethodSymbol method)
    => method != null
        && method.Name == WellKnownMemberNames.ObjectEquals
        && method.ReturnType.SpecialType == SpecialType.System_Boolean
        && method.Parameters.Length == 1
        && !method.IsImplicitlyDeclared;
```

src/NationalInstruments.Analyzers/Properties/Resources.resx
```diff
@@ -414,6 +414,16 @@
   <data name="NI1816X_Title" xml:space="preserve">
     <value>Disposable should not have finalizer</value>
   </data>
+  <data name="NI1019_Description" xml:space="preserve">
+    <value>The default equality method implementation for records is to compare enumerable types by reference, not the element values;  thus it is important for records with enumerable properties to override the default implementation and compare the enumerables using SequenceEqual or UnorderedEquals (depending on whether ordering matters).</value>
+  </data>
+  <data name="NI1019_Message" xml:space="preserve">
+    <value>Record '{0}' has properties that are enumerable collections, but does not override the default Equals and GetHashCode methods</value>
```
pakdev: Since we're only checking for `Equals`, maybe we should remove `GetHashCode` from this message. Although it'd be really weird, they could have already overrode the default implement for `GetHashCode`, right?

src/NationalInstruments.Analyzers/Properties/Resources.resx
```diff
@@ -414,6 +414,16 @@
...
+    <comment>{0} is the name of the record</comment>
+  </data>
+  <data name="NI1019_Title" xml:space="preserve">
+    <value>Records that have enumerable properties should override the default Equals and GetHashCode methods</value>
```
pakdev: Maybe we should change the verbiage from "should override the default Equals" to "explicitly implement Equals" - developers may get confused when they try to `override Equals` and find that they can't

tests/NationalInstruments.Analyzers.UnitTests/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzerTests.cs
pakdev, line 1: Could add a test that has a "fake" Equals method e.g. wrong number of parameters, wrong return type

pakdev:
Overall great job and I (we) appreciate the contribution - thanks for educating me on this issue!

pakdev (changes requested):
Overall great job and I (we) appreciate the contribution - thanks for educating me on this issue!

src/NationalInstruments.Analyzers/Correctness/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs
jonathanou: My original intention was to make sure this case does not report a diagnostic:

```cs
public record BaseRecord(IEnumerable<int> MyInts)
{
    bool virtual Equals(BaseRecord? other)
    {
        // ...
    }

    int override GetHashCode()
    {
        //...
    }
}

public record TestRecord(IEnumerable<int> MyInts)
    : BaseRecord(MyInts);
```

Here, the base record defines `MyInts`, and does correctly implement `Equals`.  The `TestRecord` derives from it, and though it does has an enumerable property when inspected, the property is derived from the base class and is not defined by `TestRecord`—so no diagnostic should be reported even if `TestRecord` does not have an `Equals` implementation.

But you're right, the code you wrote above will cause the analyzer to incorrectly suppress the diagnostic.  Let me add your code as a test case, I think we can fix it by just adding an additional check to verify the `baseTypeProperty` type is the same as the derived record's property type.

src/NationalInstruments.Analyzers/Correctness/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs
jonathanou: Sure that's definitely an improvement!  I'll add it.

src/NationalInstruments.Analyzers/Properties/Resources.resx
jonathanou: >they could have already overrode the default implement for GetHashCode, right?

I originally wanted to mention both `Equals` and `GetHashCode` to avoid the developer implementing `Equals` only to later find out from a different warning that they needed to override `GetHashCode` too.

But it's true that developers could override `GetHashCode` but not `Equals`—I discovered that you do *not* get a warning in that case—and find it a bit odd about the message.  However the relationship between `Equals` and `GetHashCode` is not very well known among less experienced developers, and if the developer knows to override `GetHashCode` already, then it's highly likely the developer would know to implement `Equals` too.

Perhaps I can reword it to `does not override *both* the default Equals and GetHashCode methods` (emphasis mine)?

src/NationalInstruments.Analyzers/Properties/Resources.resx
jonathanou: For a standalone record, yes, override does not make sense;  if the record derives from another record, however, then overriding `Equals` is the right term to use.

But I'm inclined to agree that "explicitly implement Equals" is probably better since it covers more cases, and one could interpret overriding `Equals` as "explicitly" implementing the derived `Equals` method too.

jonathanou: Ah I see why you preferred "explicitly", since the API has the notion of an "implicitly declared" equality method that is generated by the compiler.  That makes a lot more sense now given the context, and makes it more clear that "explicit implementation" is the right term to use regardless of record inheritance or not.

tests/NationalInstruments.Analyzers.UnitTests/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzerTests.cs
jonathanou: Good idea, I've added the test and updated the implementation.

src/NationalInstruments.Analyzers/Correctness/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs
jonathanou: I've added the extension methods with some slight modifications:

1.  `IsExplicitEquals` does not check for `method != null` because the code has nullable reference types enabled, so `IMethodSymbol` corresponds to a non-null type already (we get a warning if a nullable type is passed in).
2. I put `ExplicitlyImiplementsEquals` under `ITypeSymbolExtensions` because `ITypeSymbol` has the same `GetMembers` method already.

The ability to check `IsImplicitlyDeclared` was definitely new to me, and allowed me to stop checking the syntax tree manually, so thank you for the tip!

src/NationalInstruments.Analyzers/Properties/Resources.resx
jonathanou: I reworded the message, let me know if that looks good or not.

src/NationalInstruments.Analyzers/Properties/Resources.resx
jonathanou: Used "explicitly implement equality" throughout the title and description, the messages should read much clearer now.

src/NationalInstruments.Analyzers.Utilities/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -82,5 +82,18 @@ public static ImmutableArray<IMethodSymbol> GetOriginalDefinitions(this IMethodS
 
             return originalDefinitionsBuilder.ToImmutable();
         }
+
+        /// <summary>
+        /// Whether the method is the ObjectEquals method and explicitly declared
```
pakdev: Saying "whether the method is the `ObjectEquals`..." is slightly misleading because it also detects `IEquatable<T>.Equals(T)` as `WellKnownMemberNames.ObjectEquals` has a value of `Equals`

src/NationalInstruments.Analyzers.Utilities/Extensions/ITypeSymbolExtensions.cs
```diff
@@ -104,5 +107,60 @@ public static bool IsOrInheritsFromClass(this ITypeSymbol type, string className
...
+                    && m.DeclaredAccessibility == Accessibility.Public)
+                .Cast<IPropertySymbol>()
+                .ToImmutableArray();
+        }
+
+        /// <summary>
+        /// Gets whether this type is an enumerable.
+        /// </summary>
+        /// <param name="typeSymbol">the symbol to inspect.</param>
+        /// <returns>True if the type implements IEnumerable, false otherwise.</returns>
```
pakdev, line 130: Minor: your capitalization of documentation is inconsistent. While I'd prefer SA1628 to be enabled so that we're all consistent, [it's not](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1057).

src/NationalInstruments.Analyzers.Utilities/Extensions/IMethodSymbolExtensions.cs
jonathanou: Well turns out I did not intend to say "ObjectEquals", as that has the signature `Equals(Object? obj)`—I wanted to check `IEquatable<T>.Equals(T? other)`.

According to [MSDN](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#value-equality), the object equals *cannot* be declared explicitly in records, only `IEquatable<T>.Equals` can (emphasis mine):

>To implement value equality, the compiler synthesizes several methods, including:
>
>    An override of [Object.Equals(Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals#system-object-equals(system-object)). **It's an error if the override is declared explicitly**.
>
>    This method is used as the basis for the [Object.Equals(Object, Object)](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals#system-object-equals(system-object-system-object)) static method when both parameters are non-null.
>
>    A virtual, or sealed, Equals(R? other) where R is the record type. This method implements [IEquatable<T>](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1). **This method can be declared explicitly**.

So there isn't a need to look for `Equals(Object? obj)`, we only need to look for `Equals(T? other)`.

The code was already checking the parameter type is correct, I just didn't know what it was officially called until reading MSDN.  I've updated the method names and documentation to reflect this new understanding.

src/NationalInstruments.Analyzers.Utilities/Extensions/ITypeSymbolExtensions.cs
jonathanou: Capitalization updated

src/NationalInstruments.Analyzers/Correctness/RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs
jonathanou: Updated the code to check the property type as well as the name.  Added tests to cover this case as well.

pakdev: Did you push those changes?

What if `TestRecord` in your example above was defined as `public record TestRecord(IEnumerable<int> OtherInts) : BaseRecord(OtherInts)`? Users will have to implement `Equals` for `TestRecord` because the names don't match?

jonathanou: > Did you push those changes?

Yes.  It starts on line 44 in `RecordWithEnumerablesShouldOverrideDefaultEqualityAnalyzer.cs`:

```cs
var baseTypeProperties = GetBaseTypeProperties(typeSymbol)
    .ToImmutableHashSet(SymbolEqualityComparer.Default);

var enumerableProperties = typeSymbol
    .GetPublicPropertySymbols()
    .Where(p => p.Type.IsEnumerable() && !baseTypeProperties.Contains(p))
    .ToImmutableArray();
```

Instead of just tracking the property name, I'm tracking the property symbol itself (and using the symbol equality comparer, otherwise Contains() uses reference equals).

> What if TestRecord in your example above was defined as public record TestRecord(IEnumerable<int> OtherInts) : BaseRecord(OtherInts)? Users will have to implement Equals for TestRecord because the names don't match?

Yes that's correct.  If the derived record declares a differently named property, the compiler will generate a new property in the derived record,  so equality must be implemented.

You can see how the compiler generates the record in [sharplab.io](https://sharplab.io/#v2:CYLg1APgAgDABFAjAFgNwFgBQWoGY4BOApgMYD2BwCATHAEICGAzkQBRK4A8AlgHYAuAPjgBZAJ4BJAUwCUGbJjyFSFKlFoARIgW4A3IsHaIufIaMnSZWODbgh6zNuKn9Z8nPmLlKNOFp36wAAKBGQADgBKRLwMALYGRiYCwgDy/AAW2i6y1rb2jCysaZkE2XJAA).  I've posted the relevant code here: 

### Original user code

```cs
public record Base(IEnumerable<int> MyInts);

public record class Derived(IEnumerable<int> MyInts)
    : Base(MyInts);

public record DerivedPropRenamed(IEnumerable<int> OtherInts)
    : Base(OtherInts);
```

### Generated base record

```cs
public class Base : IEquatable<Base>
{
    [CompilerGenerated]
    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
    private readonly IEnumerable<int> <MyInts>k__BackingField;

    public IEnumerable<int> MyInts
    {
        [CompilerGenerated]
        get
        {
            return <MyInts>k__BackingField;
        }
        [CompilerGenerated]
        init
        {
            <MyInts>k__BackingField = value;
        }
    }

    public Base(IEnumerable<int> MyInts)
    {
        <MyInts>k__BackingField = MyInts;
        base..ctor();
    }
}
```

### Generated derived record with matching property name:

```cs
public class Derived : Base, IEquatable<Derived>
{
    public Derived(IEnumerable<int> MyInts)
        : base(MyInts)
    {
    }
}
```

Notice there is no new property declared here.

### Generated derived record with differently named property:

```cs
public class DerivedPropRenamed : Base, IEquatable<DerivedPropRenamed>
{
    [CompilerGenerated]
    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
    private readonly IEnumerable<int> <OtherInts>k__BackingField;

    public IEnumerable<int> OtherInts
    {
        [CompilerGenerated]
        get
        {
            return <OtherInts>k__BackingField;
        }
        [CompilerGenerated]
        init
        {
            <OtherInts>k__BackingField = value;
        }
    }

    public DerivedPropRenamed(IEnumerable<int> OtherInts)
    {
        <OtherInts>k__BackingField = OtherInts;
        base..ctor(OtherInts);
    }
}
```
Notice how there is a brand new property declared in the derived class.

For completeness, I'll add this as a test case.

jonathanou: Test case added.

pakdev: Cool, never heard of sharplab.io

src/NationalInstruments.Analyzers/Properties/Resources.resx
```diff
@@ -414,6 +414,16 @@
   <data name="NI1816X_Title" xml:space="preserve">
     <value>Disposable should not have finalizer</value>
   </data>
+  <data name="NI1019_Description" xml:space="preserve">
+    <value>The default implicit equality method implementation for records is to compare enumerable types by reference, not the element values;  thus it is important for records with enumerable properties to explicitly implement equality and compare the enumerables using SequenceEqual or UnorderedEquals (depending on whether ordering matters).</value>
+  </data>
+  <data name="NI1019_Message" xml:space="preserve">
+    <value>Record '{0}' has properties that are enumerable collections, but does not explicitly implement the Equals and GetHashCode methods</value>
```
pakdev: - Either remove the comma or add the subject to make it an independent clause
- I would explicitly say which Equals method because my mind would immediately think `override Equals(object other)`

"Record '{0}' has properties that are enumerable collections, but it does not explicitly implement the Equals(TRecord other) and GetHashCode methods"

src/NationalInstruments.Analyzers/Properties/Resources.resx
```diff
@@ -414,6 +414,16 @@
   <data name="NI1816X_Title" xml:space="preserve">
     <value>Disposable should not have finalizer</value>
   </data>
+  <data name="NI1019_Description" xml:space="preserve">
+    <value>The default implicit equality method implementation for records is to compare enumerable types by reference, not the element values;  thus it is important for records with enumerable properties to explicitly implement equality and compare the enumerables using SequenceEqual or UnorderedEquals (depending on whether ordering matters).</value>
```
pakdev: "Default implicit" seems redundant.

What do you think of the following?

"By default, record equality is determined by comparing property values. If one of those properties is an enumerable, though, its "value" is a reference to an object. Thus, you must explicitly implement Equals(TRecord other) and compare enumerable elements with SequenceEqual (if order matters) or UnorderedEquals for the comparison to make logical sense."

jonathanou: I think it's important to highlight that Enumerable.Equals compares the enumerable reference equality, and why that is bad.

I modified it slightly to this:

"By default, record equality is determined by calling Equals on property values. An enumerable's Equals method compares the enumerable's reference, making it a poor choice for determining whether two enumerables have the same elements . Thus, you must explicitly implement Equals(TRecord other) and compare enumerable elements with SequenceEqual (if order matters) or UnorderedEquals for the comparison to make logical sense."

What do you think?

src/NationalInstruments.Analyzers/Properties/Resources.resx
jonathanou: Sounds good!  I've updated it.

src/NationalInstruments.Analyzers/Properties/Resources.resx
```diff
@@ -414,6 +414,16 @@
   <data name="NI1816X_Title" xml:space="preserve">
     <value>Disposable should not have finalizer</value>
   </data>
+  <data name="NI1019_Description" xml:space="preserve">
+    <value>By default, record equality is determined by calling Equals on property values. An enumerable's Equals method compares the enumerable's reference, making it a poor choice for determining whether two enumerables have the same elements . Thus, you must explicitly implement Equals(TRecord other) and compare enumerable elements with SequenceEqual (if order matters) or UnorderedEquals for the comparison to make logical sense.</value>
```
pakdev:
```suggestion
    <value>By default, record equality is determined by calling Equals on property values. An enumerable's Equals method compares the enumerable's reference, making it a poor choice for determining whether two enumerables have the same elements. Thus, you must explicitly implement Equals(TRecord other) and compare enumerable elements with SequenceEqual (if order matters) or UnorderedEquals for the comparison to make logical sense.</value>
```

src/NationalInstruments.Analyzers/Properties/Resources.resx
pakdev: "poor choice" is a bit of an understatement, haha, but otherwise, I can live with it ;)

pakdev (approved):

jryckman (approved):

PR #95, opened 2024-03-19
Switch to Nuget Central Package Management

jonathanou:
# Justification
Since .NET 6.0.300, there is a new [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management) (CPM) feature that allows tracking nuget versions centrally. CPM relies on a new `Directory.Packages.props` file to maintain the list of nuget package versions and by default all projects cannot specify nuget versions—the build will break otherwise.

We should switch to using CPM to help simplify our build management, and ensure that all projects are using the same nuget package versions—switching to CPM will also make updating nuget packages in the future a lot easier.

# Implementation

- Extract all package reference versions to `Directory.Packages.props`.
- Remove version specifications in the original `<PackageReference` elements.

# Testing
Continues to build.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,28 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally Condition="'$(ManagePackageVersionsCentrally)' == ''">true</ManagePackageVersionsCentrally>
+
+    <!-- Indicates that a rebuild is required if this file changes. -->
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+  </PropertyGroup>
+
+  <ItemGroup Label="Global package reference versions">
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.2.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3" />
+    <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.2.32" />
+    <PackageVersion Include="StyleCop.Analyzers.Unstable" Version="1.2.0.435" />
+  </ItemGroup>
+
+  <ItemGroup Label="Global test package reference versions">
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.2.0" />
+    <PackageVersion Include="xunit" Version="2.4.1" />
+    <PackageVersion Include="xunit.assert" Version="2.4.1" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.1" />
+  </ItemGroup>
+
+  <ItemGroup Label="Global NuGet Package Versions">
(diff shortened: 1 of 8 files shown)


Directory.Packages.props
pakdev, line 1: Was this file handcrafted? The same file in another project just lists `<PackageVersions>`. Does that mean I'll have to manually add new entries in the future vs. use VS?

pakdev (approved):

Directory.Packages.props
jonathanou: The file was handcrafted, but VS knows how to update this file automatically when you update nuget references from the UI—you don't need to manually update the file when working in VS.

jryckman (approved):
Approving now that Pete has approved

Directory.Packages.props
jryckman: @pakdev are you good with this resolution?

pakdev: I'm guessing new `<PackageVersion>`' will not be placed in the correct `<ItemGroup>` and that we'll manually have to move them around manually?

jonathanou: I tested this in VS, and no, VS won't change the organization or structure.  The only thing it does is remove blank spaces between item groups, but that's not a big deal I think.

image

pakdev: After trying it myself, I'm guessing VS inserts each new `<PackageVersion>` in the next alphabetical slot. Thus, everything will get added to "Global package reference versions". It'd be less of a brain burden to just not have sections, IMO. As it is, I don't know what the difference between "Global package reference versions" and "Global NuGet Package Versions" is. Thoughts @jryckman ?

Directory.Packages.props
```diff
@@ -0,0 +1,28 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally Condition="'$(ManagePackageVersionsCentrally)' == ''">true</ManagePackageVersionsCentrally>
+
+    <!-- Indicates that a rebuild is required if this file changes. -->
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
```
pakdev: I think this happens [automatically](https://github.com/dotnet/sdk/issues/277#issuecomment-257720464), but I haven't found the exact location to prove it

pakdev: https://www.panopticoncentral.net/2019/07/12/msbuildallprojects-considered-harmful/

Directory.Packages.props
jonathanou: The sections are not important to me, so if you like no sections, I can do that.

I originally used sections to help improve readability.  I did not anticipate we would be adding new nuget package references very often given that this codebase sees very little development activity.

jonathanou: Sections removed.

Directory.Packages.props
jonathanou: OK I've removed it.

Directory.Packages.props
pakdev: Sorry, arguments could go either way, but I appreciate you removing the sections. Don't let my nitpicky feedback belie my appreciation for your work

PR #96, opened 2024-03-20
Update code analysis and xunit nuget versions

jonathanou:
# Justification
The CSharp analyzer has quite a few updates since 4.2.0, including new APIs and new analyzer rules for writing analyzers.

To help keep the analyzer code current, we should upgrade the analyzer versions and fix any new warnings in our analyzers.

# Implementation
- Update the nuget packages related to implementing analyzers to latest stable—the code analyzer nugets, such as style cop, are *not* updated.
- Fix any new warnings in our custom analyzer code.

# Testing
Code builds and all tests continue to pass.

### DIFF
--- a/src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceCodeFixProvider.cs
+++ b/src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceCodeFixProvider.cs
@@ -1,7 +1,7 @@
-using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Composition;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -84,6 +84,10 @@ namespace NationalInstruments.Analyzers.Correctness
                 return Task.FromResult(_context.Document);
             }
 
+            [SuppressMessage(
+                "MicrosoftCodeAnalysisCorrectness",
+                "RS1035:The symbol 'File' is banned for use by analyzers: Do not do file IO in analyzers",
+                Justification = "Existing working code.")]
             private void ApproveNamespace(string namespaceName, string namespacesFilePath)
             {
                 var lines = File.ReadAllLines(namespacesFilePath);
--- a/src/NationalInstruments.Analyzers/Correctness/AwaitInReadLockOrTransactionAnalyzer.cs
+++ b/src/NationalInstruments.Analyzers/Correctness/AwaitInReadLockOrTransactionAnalyzer.cs
@@ -1,4 +1,5 @@
 using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Linq;
 using Microsoft.CodeAnalysis;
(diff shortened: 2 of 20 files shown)


src/NationalInstruments.Analyzers/Correctness/AwaitInReadLockOrTransactionAnalyzer.cs
```diff
@@ -16,7 +17,11 @@ public sealed class AwaitInReadLockOrTransactionAnalyzer : NIDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "NI1015";
 
-        public static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+        [SuppressMessage(
+            "MicrosoftCodeAnalysisCorrectness",
+            "RS1035:'CultureInfo.CurrentCulture' is banned for use by analyzers: Analyzers should use the locale given by the compiler command line arguments, not the CurrentCulture",
+            Justification = "No public API available to get the locale given to the compiler, see https://github.com/dotnet/roslyn/issues/66566")]
```
jryckman, line 23: This justification looks good to as to why it should be suppressed, thanks!

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceCodeFixProvider.cs
```diff
@@ -84,6 +84,10 @@ protected override Task<Document> GetChangedDocumentAsync(CancellationToken canc
                 return Task.FromResult(_context.Document);
             }
 
+            [SuppressMessage(
```
jryckman, line 87: Is there an easy fix to address this warning rather than suppressing it?

src/NationalInstruments.Analyzers/NationalInstruments.Analyzers.csproj
```diff
@@ -5,6 +5,7 @@
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
     <AssemblyName>NationalInstruments.Analyzers</AssemblyName>
     <RootNamespace>NationalInstruments.Analyzers</RootNamespace>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
```
jryckman, line 8: What does this attribute do?

jryckman (approved):
Thank you for taking the time to keep us current! I left one open question for you.

src/NationalInstruments.Analyzers/NationalInstruments.Analyzers.csproj
jonathanou: This attribute enables additional analyzer rules.  There's actually a warning if you don't enable this.

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceCodeFixProvider.cs
jonathanou: I'm not exactly sure, it seems there's no way around writing the approved namespace file.

Also, I think the warning is overly aggressive because this is a code fix provider, not an analyzer—I think it shouldn't even participate in compilation?  The whole point of banning file IO during compilation is because it is tricky during incremental build.  Code fix providers shouldn't affect building (as far as I understand it).  

Separately, the warning that this code fix provider is fixing, NI1800, is not enabled by default.  In ASW we also disable it with this comment:

```ini
# Missing the namespace approval file - we don't do this namespace analysis with analyzers yet.
dotnet_diagnostic.NI1800.severity = none
```

So it seems this rule and code fix provider isn't really being used currently.

src/NationalInstruments.Analyzers/NationalInstruments.Analyzers.csproj
jryckman: Sounds good

src/NationalInstruments.Analyzers/Correctness/ApprovedNamespaceCodeFixProvider.cs
jryckman: Okay. @pakdev are you okay with this warning in this analyzer? Is this indicative of another issue that we need to resolve "soon" ?

pakdev: Not worried about this for the reasons @jonathanou stated

pakdev (approved):

PR #100, opened 2024-11-21
Update net analyzers to 9.0

jonathanou:
# Justification
With the release of .NET 9 there are new analyzers we can use.

# Implementation
- Update `Microsoft.CodeAnalysis.NetAnalyzers` to `9.0.0`.
- Update the default `AnalysisLevel` to `9.0` so that consumers using older .NET SDK will still get to use the new .NET 9 analyzers.

# Testing
Built nupkg locally and confirmed new warnings are emitted in ASW when the analyzer package is updated.

PR #101, opened 2025-11-18
Update net analyzers to 10.0

yaqian256:
# Justification
With the release of .NET 10 there are new analyzers we can use.

# Implementation
- Address the feedback from https://github.com/ni/csharp-styleguide/pull/102
- Update Microsoft.CodeAnalysis.NetAnalyzers to 10.0.100.
- Update the default AnalysisLevel to 10.0 so that consumers using older .NET SDK will still get to use the new .NET 10 analyzers.

# Testing
Built nupkg locally and confirmed new warnings are emitted in my projects when the analyzer package is updated.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -5,6 +5,7 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
+    <!-- `clear` ensures no additional sources are inherited from another config file. -->
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
--- a/nuget.config
+++ b/nuget.config
@@ -5,5 +5,7 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/nuget.config
+++ b/nuget.config
@@ -5,6 +5,5 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
-    <add key="NuGet" value="https://www.nuget.org/api/v2/" />
   </packageSources>
 </configuration>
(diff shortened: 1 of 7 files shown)


nuget.config
```diff
@@ -5,5 +5,8 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
+    <!-- `clear` ensures no additional sources are inherited from another config file. -->
```
yaqian256, line 8: NOTE: This is to adddress feedback in https://github.com/ni/csharp-styleguide/pull/102

yaqian256:
NOTE: followed https://github.com/ni/csharp-styleguide/pull/100 to make this change.

src/AnalyzerConfiguration/NI.CSharp.Analyzers.targets
```diff
@@ -48,7 +48,7 @@
     <Analyzer Include="$(PkgStyleCop_Analyzers_Unstable)\analyzers\dotnet\cs\StyleCop.Analyzers.CodeFixes.dll"/>
 
     <!-- Microsoft's CA analyzers -->
-    <Analyzer Include="$(PkgMicrosoft_CodeAnalysis_NetAnalyzers)\analyzers\dotnet\cs\Microsoft.CodeAnalysis.NetAnalyzers.dll"/>
+    <Analyzer Include="$(PkgMicrosoft_CodeAnalysis_NetAnalyzers)\analyzers\dotnet\Microsoft.CodeAnalysis.NetAnalyzers.dll"/>
```
yaqian256, line 51: NOTE: The `Microsoft.CodeAnalysis.NetAnalyzers.dll` gets moved to the upper folder in 10.0. See https://nuget.info/packages/Microsoft.CodeAnalysis.NetAnalyzers/10.0.100

nick-beer (approved):

jryckman (approved):

PR #102, opened 2025-11-18
Update to .NET 8

yaqian256:
# Justification
.NET 6 has reached its end of life, and the `actions/setup-dotnet@v1` step used in the csharp-styleguide workflow no longer installs it. We need to update to .NET 8 to prevent build errors.

# Implementation
- Updated global.json to use .NET 8 (same as the one in ASW)
- Removed the NuGet package source in nuget.config to void NU1507 warnings
- Updated the NIExecutableTargetFramework and NILibraryTargetFramework properties to net8.0.

# Testing
builds

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -5,6 +5,5 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
-    <add key="NuGet" value="https://www.nuget.org/api/v2/" />
   </packageSources>
 </configuration>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,8 +8,8 @@
     </PropertyGroup>
 
     <PropertyGroup>
-        <NILibraryTargetFramework>net6.0</NILibraryTargetFramework>
-        <NIExecutableTargetFramework>net6.0</NIExecutableTargetFramework>
+        <NILibraryTargetFramework>net8.0</NILibraryTargetFramework>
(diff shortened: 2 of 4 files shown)


jryckman:
@nick-beer @pakdev is there anyone else using .NET right now mainly in ASW that can also chime in on this change?

nuget.config
```diff
@@ -5,6 +5,5 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
-    <add key="NuGet" value="https://www.nuget.org/api/v2/" />
```
yaqian256, line 8: NOTE: Removed this to avoid warning NU1507: There are 2 package sources defined in your configuration. When using central package management, please
map your package sources with package source mapping (https://aka.ms/nuget-package-source-mapping) or specify a single package source. The following sources are defined: NuGet, nuget.org

nick-beer: I'm not sure I understand this change? I think it would be better to add a `<clear />` here.

nick-beer (approved):

jryckman (approved):

jryckman:
@yaqian256 The change was approved and the PR went in however, please address Nick's comment in nuget.config before I build a new export.

yaqian256:
> @yaqian256 The change was approved and the PR went in however, please address Nick's comment in nuget.config before I build a new export.

addressed in https://github.com/ni/csharp-styleguide/pull/101

PR #104, opened 2025-12-04
Fix NuGet publish path after .NET 8 upgrade

yaqian256:
# Justification
The NuGet publishing pipeline failed because the built NuGet package could not be located. This issue occurred after upgrading to .NET 8 (see #102), as the build configuration is now included in the NuGet path.

image

# Implementation
Update the path to the nupkg to include `Release` 

# Testing
Verified the new path exists.


jryckman:
Thanks for looking into this failure. I wasn't able to look into it yet. I will merge your changes in and see where things are. There are changes that were rolled back so we need to ensure that all of your changes are still in there

jryckman (approved):

